PR #7, opened 2022-01-18
Local/Remote Source Support & User Federation Implementation

motaparthipavankumar:
Local path & Source URL installation support:

    Changes to pass local keycloak zip file.
    Changes to pass local RHSSO zip file.
    Changes to pass source url to download RHSSO zip file from.

Now all following scenarios are available:

    RHSSO from CSP
    RHSSO from remote source
    RHSSO from localpath
    Keycloak from remote source
    Keycloak from local path

Added User Federation changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,19 +31,54 @@ collections:
 
 ### Install Playbook
 
-`playbooks/keycloak.yml` installs the keycloak or Red Hat Single Sign-On based on the defined variables.
+`playbooks/keycloak.yml` installs the upstream(Keycloak) based on the defined variables.
+`playbooks/rhsso.yml` installs Red Hat Single Sign-On(RHSSO) based on defined variables.
 
-### Choosing between Red Hat products and upstream project
+### Choosing between upstream(Keycloak) project and Red Hat Single Sign-On(RHSSO)
 
-The roles supports installing Red Hat Single Sign-On from the Customer Portal, when the following variables are defined:
+The roles supports installing upstream(Keycloak) or Red Hat Single Sign-On in the following ways
+
+#### Install upstream(Keycloak) from remote source
+
+This is default approach, there is one required variable
 
 ```
+keycloak_admin_password: "<changeme>"
+```
+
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -31,20 +31,55 @@ collections:
...
 ```
 
 where `sso_product_id` is the ID for the specific Red Hat Single Sign-On version, ie. _101971_ will install version _7.5_)
 
+#### Install RHSSO from remote sources like Nexus etc, when the following variables are defined
+
+```
+keycloak_admin_password: "<changeme>"
+keycloak_rhsso_enable: True
+rhsso_source_download_url: '<url to download RHSSO zip file>'
```
guidograzioli, line 72: we could reuse `keycloak_rhsso_base_url` ^

roles/keycloak/README.md
```diff
@@ -75,20 +76,44 @@ The following variables are _required_ only when `keycloak_db_enabled` is True:
...
 
+The following variable can be used to install Keycloak or Red Hat Single Sign-On from local path:
+| Variable | Description | Example |
+|:---------|:------------|:---------|
+|`zip_file_local_path` | Full local path of upstream(Keycloak) or Red Hat Single Sign-On zip file on Ansible control plane | `tmp/rhsso/rh-sso-7.5-server-dist.zip` |
+
+The following variable can be used to install Red Hat Single Sign-On from source via url, auth support is not added right now.
+| Variable | Description | Example |
+|:---------|:------------|:---------|
+|`rhsso_source_download_url` | URL to download Red Hat Single Sign-On zip file from source | `http://localhost:8081/nexus/rhsso/rh-sso-7.5-server-dist.zip` |
```
guidograzioli, line 88: The idea here would be to always rely on a local file for installation; from where, three execution paths open: 
1) download from rhn to that temp file, then distribute the file to nodes
2) download from alternate uri to that temp file, then distribute the file to nodes
3) the file is there and we are offline, just distribute

I'll forward fix before a new release

PR #8, opened 2022-01-19
Client Creation Enhancement

motaparthipavankumar:
Create clients enhancement, to support more parameters.


roles/keycloak_realm/tasks/manage_client_roles.yml
```diff
@@ -3,10 +3,12 @@
     name: "{{ item }}"
     realm: "{{ client.realm }}"
     client_id: "{{ client.name }}"
+    description: "{{ item.description | default(omit) }}"
```
guidograzioli, line 6: this was meant to be `client.description` ? `roles` is a list of strings, if you want to allow name and description for roles, it has to be modified and documented as list of dicts; see: https://github.com/ansible-middleware/keycloak/blob/0fdd89d571bc5bbbd2998884563c94bef640d544/playbooks/keycloak-realm.yml#L53
and then the above become name: "{{ item.name }}" and description: "{{ item.description | default(omit) }}"

roles/keycloak_realm/tasks/main.yml
```diff
@@ -34,42 +34,35 @@
...
+- name: Create or update a Keycloak client
   community.general.keycloak_client:
     auth_client_id: "{{ keycloak_auth_client }}"
     auth_keycloak_url: "{{ keycloak_url }}/auth"
     auth_realm: "{{ keycloak_auth_realm }}"
     auth_username: "{{ keycloak_admin_user }}"
     auth_password: "{{ keycloak_admin_password }}"
-    client_id: "{{ item.name }}"
     realm: "{{ item.realm }}"
-    default_roles: "{{ item.roles | default(omit) }}"
```
guidograzioli, line 64: at this point better to keep this, because it is mapped from keycloak API (rhsso does not honor the field, so the next task to create the roles is still necessary)

guidograzioli (approved):
LGTM

PR #18, opened 2022-03-08
Change to allow installation of latest RHSSO patch

### LINKED ISSUE #14: Support update to patch version of RH SSO
Currently the installation of RH-SSO via the keycloak role is only possible on baseline versions;
dedicated tasks are needed to perform cluster patch updates

guidograzioli:
- add tasks for running cli
- add tasks and config for downloading latest patch, and apply via cli

Fixes #14

### DIFF
--- a/roles/keycloak/tasks/rhsso_cli.yml
+++ b/roles/keycloak/tasks/rhsso_cli.yml
@@ -6,15 +6,8 @@
     fail_msg: "Missing required parameters to execute CLI."
     quiet: true
 
-- name: Verify server management interface is functional
-  ansible.builtin.uri:
-    url: "{{ keycloak_management_url }}"
-  register: result
-  until: result.status == 200
-  retries: 5
-  delay: 5
-
 - name: "Execute CLI query: {{ query }}"
   ansible.builtin.command: >
-    {{ path_to_cli }} -c --output-json --command='{{ query }}' --controller={{ keycloak_host }}:{{ keycloak_management_http_port }}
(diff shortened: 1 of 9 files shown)


roles/keycloak/tasks/rhsso_cli.yml
```diff
@@ -0,0 +1,20 @@
...
+
+- name: Verify server management interface is functional
+  ansible.builtin.uri:
+    url: "{{ keycloak_management_url }}"
+  register: result
+  until: result.status == 200
+  retries: 5
+  delay: 5
+
+- name: "Execute CLI query: {{ query }}"
```
rpelisse, line 17: @guidograzioli as this collection depends on wilfly, will not reuse the jboss_cli.yml from it?

rpelisse: I see now that you are using the --out-json. I think we actually need to enhance jboss_cli to use it too and returns directly json.

rpelisse (approved):

roles/keycloak/tasks/rhsso_cli.yml
guidograzioli: That was the initial idea of importing the eap role, but no matter what i tried, the jboss_cli distributed with keycloak/rhsso does not support to spit out json, so I am in the process of parsing DMR instead

rpelisse (approved):

rpelisse:
@guidograzioli LGTM!

PR #19, opened 2022-03-11
misc: simplify fastpackages logic and reduce play time

guidograzioli:


rpelisse:
@guidograzioli don't forget to change the comment from chore: to misc: !

PR #26, opened 2022-03-24
move admin pwd param from vars to defaults, add assert

### LINKED ISSUE #22: Setting administrator password is too clumsy
##### SUMMARY

The `keycloak_admin_password` is declared empty in the keycloak role varfile. Given [variable precedence](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#understanding-variable-precedence) it becomes only possible to set the value in include_task  parameters or commandline. Find a better solution

##### ISSUE TYPE
 - Bug Report

##### COLLECTION VERSION

```
1.0.0
```

##### STEPS TO REPRODUCE

```yaml
- playbook: [..]
  vars:
    keycloak_admin_password: "changeme"
  roles:
     -  keycloak
```

##### EXPECTED RESULTS

keycloak_admin_password is overriden by playbook vars, or by inventory host|group_vars

##### ACTUAL RESULTS

It is only possible to override with extra_vars, or include_role vars

```

```

guidograzioli:
Fix #22

### DIFF
--- a/molecule/default/verify.yml
+++ b/molecule/default/verify.yml
@@ -8,3 +8,4 @@
       ansible.builtin.assert:
         that:
           - ansible_facts.services["keycloak.service"]["state"] == "running"
+          - ansible_facts.services["keycloak.service"]["status"] == "enabled"
--- a/playbooks/keycloak_realm.yml
+++ b/playbooks/keycloak_realm.yml
@@ -6,7 +6,7 @@
       ansible.builtin.include_role:
         name: middleware_automation.keycloak.keycloak_realm
       vars:
-        keycloak_admin_password: "changeme"
+        keycloak_admin_password: "remembertochangeme"
         keycloak_realm: TestRealm
         keycloak_user_federation:
           - realm: TestRealm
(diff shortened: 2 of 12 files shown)


molecule/default/verify.yml
```diff
@@ -8,3 +8,4 @@
       ansible.builtin.assert:
         that:
           - ansible_facts.services["keycloak.service"]["state"] == "running"
+          - ansible_facts.services["keycloak.service"]["status"] == "enabled"
```
rpelisse, line 11: We need to add that in the validation role in the other collection!!! (also we need to turn this verify into a validation role for keycloak)

playbooks/keycloak_realm.yml
```diff
@@ -6,7 +6,7 @@
       ansible.builtin.include_role:
         name: middleware_automation.keycloak.keycloak_realm
       vars:
-        keycloak_admin_password: "changeme"
+        keycloak_admin_password: "remembertochangeme"
```
rpelisse, line 9: :)

rpelisse (approved):

PR #28, opened 2022-04-05
Switch playbooks hosts to all

rpelisse:


rpelisse:
@guidograzioli if you don't mind, I would like to use "all" in those playbooks, as it will help reusing them in other automation.

PR #30, opened 2022-04-12
Add `keycloak_config_override_template` parameter for passing a custom xml config template

### LINKED ISSUE #23: Support passing-in standalone.xml template
##### SUMMARY

Templating of configuration xml cannot cover everything, so allow to pass a custom template to the role (completely bypassing internal templating). Of course passing a preconfigured configuration means there is no guarantee variables for the role and what is in the template matches, the documentation must clearly state that.

##### ISSUE TYPE
 - Feature Idea

guidograzioli:
Fix #23


rpelisse (approved):
LGTM!

PR #38, opened 2022-06-30
Use become for tasks that will otherwise fail

cattz:
The modified tasks require sudo rights to be performed.
Running the whole playbook as sudo will fail on the tasks that are delegated to localhosts (unless you are actually running the playbook as root)


guidograzioli:
Thanks for the PR, I'll merge after ci completes

guidograzioli (approved):

PR #39, opened 2022-06-30
keycloak_quarkus: use absolute path for certificate files

cattz:
If you want to use a certificate for both Keycloak and Nginx, the cert files are quite unlikely to be under `keycloak.home`
Even in case you want to keep the under Keycloak path, you can always define the variable:

```
   keycloak_quarkus_cert_file: "{{ keycloak.home }}/conf/my_keycloak_cert.pem"
   keycloak_quarkus_key_file: "{{ keycloak.home }}/conf/my_keycloak_key.pem"
```
This change will break your previous configuration (unless you use the default values).
To fix your config for the new changes, you will have use full paths for the 2 variables mentioned above. For example:
```
   keycloak_quarkus_key_file: "/etc/ssl/certs/my_keycloak_key.pem"
   keycloak_quarkus_cert_file: "/etc/ssl/certs/my_keycloak_cert.pem"
```

### DIFF
--- a/molecule/quarkus/converge.yml
+++ b/molecule/quarkus/converge.yml
@@ -9,8 +9,8 @@
     keycloak_quarkus_http_relative_path: ''
     keycloak_quarkus_log: file
     keycloak_quarkus_https_enabled: True
-    keycloak_quarkus_key_file: conf/key.pem
-    keycloak_quarkus_cert_file: conf/cert.pem
+    keycloak_quarkus_key_file: "{{ keycloak.home }}/conf/key.pem"
+    keycloak_quarkus_cert_file: "{{ keycloak.home }}/conf/cert.pem"
   roles:
     - role: keycloak_quarkus
     - role: keycloak_realm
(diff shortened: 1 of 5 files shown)


guidograzioli:
I am ok with the change, but you'll need to update meta/argument_specs, defaults/main.yml and README as well

cattz:
This change will break existing configurations, unless they are using the default value. Not sure how to warn users about this

cattz:
I'm not sure on why CI is failing now

guidograzioli:
Found the issue here: 
https://github.com/ansible-middleware/keycloak/blob/main/molecule/quarkus/converge.yml#L12
(the molecule test playbook was not using default and broke). 
The certs are created in the prepare step here:
https://github.com/ansible-middleware/keycloak/blob/main/molecule/quarkus/prepare.yml#L9
Please amend the test config so it passes

guidograzioli:
LGTM (again) thanks for contributing

PR #41, opened 2022-08-02
Skips part of the verify.yml that cannot work on Olympus (if HERA_HOME is defined)

rpelisse:


molecule/quarkus/prepare.yml
```diff
@@ -1,19 +1,25 @@
...
     - command: openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -sha256 -days 365 -nodes -subj '/CN=instance'
       delegate_to: localhost
+
     - lineinfile:
         dest: /etc/hosts
         line: "127.0.0.1 instance"
         state: present
       delegate_to: localhost
       become: yes
+      when:
```
rpelisse, line 19: @guidograzioli not the nicest way to figure out when we run on github or olympus, but that works and will do for now if you can leav with it :)

rpelisse:
@guidograzioli after exploring a lot of silly options, I finaly realized that the test failing cannot work on Olympus as it is. The container running molecule has no network access to the 'instance'. So we need to change the way podman is used on Olympus to have network and communication between the molecule runner and its slaves (and between slaves). 
In the meantime, this approach allows to run the scenario on both gh actions and olympus. Only caveat, this one extra test in the verify.yml is not performed on Olympus...

PR #47, opened 2022-12-15
allows user to switch no_log to True for debugging purpose

rpelisse:
@guidograzioli WDYT? I'm pretty user will need to switch no_log to False for debugging purpose, so rather than having them edit the installed collection, this seems to be a better approach...

### DIFF
--- a/roles/keycloak/defaults/main.yml
+++ b/roles/keycloak/defaults/main.yml
@@ -2,7 +2,7 @@
 ### Configuration specific to keycloak
 keycloak_version: 15.0.2
 keycloak_archive: "keycloak-{{ keycloak_version }}.zip"
-keycloak_download_url: "https://github.com/keycloak/keycloak/releases/download/{{ keycloak_version }}/{{ keycloak_archive }}"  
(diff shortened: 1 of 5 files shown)


roles/keycloak/meta/argument_specs.yml
```diff
@@ -266,6 +266,10 @@ argument_specs:
                 default: "Keycloak"
                 description: "systemd description for keycloak"
                 type: "str"
+            keycloak_no_log:
+                default: false
```
guidograzioli, line 270: this should be true by default (ie. if not overridden, hide logs)

roles/keycloak/defaults/main.yml
```diff
@@ -83,3 +83,5 @@ keycloak_default_jdbc:
   mariadb:
     url: 'jdbc:mariadb://localhost:3306/keycloak'
     version: 2.7.4
+# role specific vars
+keycloak_no_log: False
```
guidograzioli, line 87: this also True (see other comment)

guidograzioli (commented):
looks good, but pls set the default value to True and add a line in README with the new variable

PR #48, opened 2022-12-16
Runner branch

rpelisse:
@guidograzioli a few little tweaks, that I did as I was debugging the ansible-runner job.

### DIFF
--- a/roles/keycloak/tasks/start_keycloak.yml
+++ b/roles/keycloak/tasks/start_keycloak.yml
@@ -11,5 +11,5 @@
     url: "{{ keycloak.health_url }}"
   register: keycloak_status
   until: keycloak_status.status == 200
-  retries: 25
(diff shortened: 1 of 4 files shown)


roles/keycloak/tasks/start_keycloak.yml
```diff
@@ -11,5 +11,5 @@
     url: "{{ keycloak.health_url }}"
   register: keycloak_status
   until: keycloak_status.status == 200
-  retries: 25
-  delay: 10
\ No newline at end of file
+  retries: "{{ keycloak_service_start_retries }}"
+  delay: "{{ keycloak_service_start_delay }}"
```
rpelisse, line 15: @guidograzioli I suspect users will need to tweak that, if nothing else, because of slow (or very fast system)

PR #49, opened 2022-12-16
Runner job

rpelisse:


rpelisse:
@guidograzioli I think there is a again some weird issue on GH action infrastructure:
```
ERROR! couldn't resolve module/action 'community.docker.docker_container'. This often indicates a misspelling, missing collection, or incorrect module path.
  
  The error appears to be in '/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/molecule_docker/playbooks/destroy.yml': line 8, column 7, but may
  be elsewhere in the file depending on the exact syntax problem.
  
  The offending line appears to be:
  
    tasks:
      - name: Destroy molecule instance(s)
        ^ here
```

PR #51, opened 2022-12-28
Allow overrides for service start retries and delay

rpelisse:
A small tweak, but I suspect many users we'll need to fine tune this depending on their env

### DIFF
--- a/roles/keycloak/meta/argument_specs.yml
+++ b/roles/keycloak/meta/argument_specs.yml
@@ -266,6 +266,14 @@ argument_specs:
                 default: "Keycloak"
                 description: "systemd description for keycloak"
                 type: "str"
+            keycloak_service_start_delay:
+                default: "10"
+                description: "Expected delay in ms before the service is expected to be available after start."
+                type: "int"
+            keycloak_service_start_retries:
+                default: "25"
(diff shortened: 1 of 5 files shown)


roles/keycloak/meta/argument_specs.yml
```diff
@@ -266,6 +266,14 @@ argument_specs:
                 default: "Keycloak"
                 description: "systemd description for keycloak"
                 type: "str"
+            keycloak_service_start_delay:
+                default: "10"
+                description: "Expected delay in ms before the service is expected to be available after start."
```
guidograzioli, line 271: ms -> seconds

roles/keycloak/tasks/start_keycloak.yml
```diff
@@ -11,5 +11,5 @@
     url: "{{ keycloak.health_url }}"
   register: keycloak_status
   until: keycloak_status.status == 200
-  retries: 25
-  delay: 10
\ No newline at end of file
+  retries: "{{ keycloak_service_start_retries }}"
```
guidograzioli, line 14: please add same vars to restart_keycloak.yml too

guidograzioli (approved):
:+1: :+1:

PR #53, opened 2023-01-26
Update keycloak.conf.j2

bbarun:
The documentation states that the path is relative to the /conf directory, which is true. So if one changes the contents of this file it would have no effect because the path would be non-existent. To make things worse, there would be no obvious error stating this and the configuration would be loaded from the defaults.


guidograzioli (approved):
LGTM

guidograzioli:
Thanks for the contribution! (and sorry it took so long to merge)

PR #88, opened 2023-05-30
Improve service restart behavior configuration

hwo-wd:
... by allowing the user to freely choose from all [available values supported by systemd](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=):

* `no`
* `always`
* `on-success`
* `on-failure`
* `on-abnormal`
* `on-abort`
* `on-watchdog`


guidograzioli:
Ok, I see; how about this:

- `keycloak_service_restart_always`: new parameter, default `False`. Used to set the systemd restart value `always`
- `keycloak_service_restart_on_failure` remains as it is, default `False`, unless `keycloak_service_restart_always` is true, in which case by default is True. Sets the value as `on-failure` when true

?

hwo-wd:
This should do the trick

guidograzioli (approved):
LGTM

PR #89, opened 2023-05-31
Change xa_datasource_class value for mariadb jdbc configuration

schmaxit:
`org.mariadb.jdbc.MySQLDataSource` is [deprecated since driver version 1.3.1](https://javadoc.io/doc/org.mariadb.jdbc/mariadb-java-client/1.3.1/org/mariadb/jdbc/MySQLDataSource.html) and has been removed on 3.x.

Since the introduction of `org.mariadb.jdbc.MariaDbDataSource`, `org.mariadb.jdbc.MySQLDataSource` has always been just an alias.

A better value for this key should be `org.mariadb.jdbc.MariaDbDataSource`.


guidograzioli:
Thanks for your contribution

PR #91, opened 2023-06-09
Undefine `keycloak_db_valid_conn_sql` default

schmaxit:
Setting an empty string default for `keycloak_db_valid_conn_sql` results in an empty value in the configuration file. We should rely on defaults set fo `keycloak_jdbc.[database].validate_query`.

The alternative cloud be using `"{{ keycloak_db_valid_conn_sql | default('select 1', true) }}"` value for `keycloak_jdbc.[database].validate_query`


guidograzioli:
Hello, thanks for the PR; what version of ansible are you on? We get the variable undefined, not empty string, in out CI testing. Anyway, I think the change makes sense, but I'd prefer the alternative `"{{ keycloak_db_valid_conn_sql | default('select 1', true) }}"`

Your thoughts?

PR #98, opened 2023-06-23
Fix #97 - proper checks for keycloak_jgroups_subnet

### LINKED ISSUE #97: #93 keycloak_jgroups_subnet introduced issue
The changes introduced via PR #93  around 
https://github.com/ansible-middleware/keycloak/blob/cebec9c717b6fc55402d9098d53236de15f3b164/roles/keycloak/templates/standalone-ha.xml.j2#L664:

```xml
        <interface name="jgroups">
{% if keycloak_jgroups_subnet is defined and keycloak_jgroups_subnet | string | length > 0 %}
            <subnet-match value="{{ keycloak_jgroups_subnet | string }}"/>
{% elif ansible_default_ipv4 is defined and (ansible_default_ipv4.network + '/' + ansible_default_ipv4.netmask) | ansible.utils.ipaddr('net') | length > 0 %}
            <subnet-match value="{{ (ansible_default_ipv4.network + '/' + ansible_default_ipv4.netmask) | ansible.utils.ipaddr('net') }}"/>
{% else %}
            <any-address />
{% endif %}
        </interface>
```

by default:
* `keycloak_jgroups_subnet` is defined (=null which is mapped to None on jinja, or sth. similar, but the end result is a string of length 4 with content `None`)
* thus the first branch is taken
* and a `<subnet-match value="None"/>`

hwo-wd:


guidograzioli (approved):
Cool, thanks!

PR #132, opened 2023-11-29
Add prefix check for keycloak_quarkus_http_relative_path

saadsb20:


saadsb20:
In my case It wasn't set to '/' ... I added an assert for that to validate that the relative path begins with '/'. 
Thanks for your time.

PR #154, opened 2024-01-10
fix/#153: keycloak_quarkus: Use `keycloak_quarkus_java_opts`

hwo-wd:
Note: when multiple -X options of the same kind are provided, the last option seems to take precendence as per <https://stackoverflow.com/a/26727332>:

> java -Xmx1G -XX:+PrintFlagsFinal -Xmx2G 2>/dev/null | grep MaxHeapSize


guidograzioli:
Thanks for your contribution!

PR #163, opened 2024-01-22
keycloak_quarkus: `sticky-session`s for infinispan routes

hwo-wd:
New extremely long unimaginatively named parameter (do not blame the collection):

|`keycloak_quarkus_spi_sticky_session_encoder_infinispan_should_attach_route`| If the route should be attached to cookies to reflect the node that owns a particular session. If false, route is not attached to cookies and we rely on the session affinity capabilities from reverse proxy | `True` |

allows to configure the corresponding option.

Fix #163

### DIFF
--- a/roles/keycloak_quarkus/templates/keycloak.conf.j2
+++ b/roles/keycloak_quarkus/templates/keycloak.conf.j2
@@ -55,8 +55,7 @@ cache-stack=tcp
 # Proxy
 proxy={{ keycloak_quarkus_proxy_mode }}
 {% endif %}
-# Do not attach route to cookies and rely on the session affinity capabilities from reverse proxy
-#spi-sticky-session-encoder-infinispan-should-attach-route=false
+spi-sticky-session-encoder-infinispan-should-attach-route={{ keycloak_quarkus_spi_sticky_session_encoder_infinispan_should_attach_route | d(true) | lower }}
 
 # Transaction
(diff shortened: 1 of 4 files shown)


roles/keycloak_quarkus/templates/keycloak.conf.j2
```diff
@@ -55,8 +55,7 @@ cache-stack=tcp
 # Proxy
 proxy={{ keycloak_quarkus_proxy_mode }}
 {% endif %}
-# Do not attach route to cookies and rely on the session affinity capabilities from reverse proxy
-#spi-sticky-session-encoder-infinispan-should-attach-route=false
+spi-sticky-session-encoder-infinispan-should-attach-route={{ keycloak_quarkus_spi_sticky_session_encoder_infinispan_should_attach_route | d(true) }}
```
guidograzioli, line 58: Just a quick q: I see this declared as a boolean, doesn't it get emitted with uppercase? (I remember having to use `| lower` for other booleans)

guidograzioli: otherwise, lgtm

hwo-wd: fixed (force pushed)

guidograzioli (approved):
:+1:

PR #178, opened 2024-03-17
Ubuntu compatibility

Aeyk:


guidograzioli (approved):
LGTM

larueli:
Thanks a lot for this contribution, I used a little bit of hacking to make this work on Ubuntu before your contribution. I have two questions.

1) Sysconfig vs default

Ubuntu does not have a `/etc/sysconfig` folder, so in the [systemd task](https://github.com/ansible-middleware/keycloak/blob/7a0a99a31c2fcd0e218d415cda5613de4d47be1e/roles/keycloak_quarkus/tasks/systemd.yml#L29), you write down the environment file to `/etc/default`.

But, in the [systemd unit file](https://github.com/ansible-middleware/keycloak/blob/7a0a99a31c2fcd0e218d415cda5613de4d47be1e/roles/keycloak_quarkus/templates/keycloak.service.j2#L8) we still reference the old /etc/sysconfig folder.

Shouldn't those lines be :

```
{% if ansible_facts.os_family == "Debian" %}
EnvironmentFile=-/etc/default/keycloak
{% endif %}
{% if ansible_facts.os_family == "RedHat" %}
EnvironmentFile=-/etc/sysconfig/keycloak
{% endif %}
```

2) keycloak_configure_iptables should be set to false in the defaults.yml, like firewalld. Should the variable be renamed as keycloak_quarkus_configure_iptables to include the full name of the role (like firewalld) ?

Tell me if I'm wrong ? I can create an other PR if you want.

guidograzioli:
It's correct; now that the PR is merged, I am doing right now some refactoring and adding a molecule test on an ubuntu container. Please check back soon or you are welcome to open any other ubuntu changeset not overlapping with the above

guidograzioli:
#181 

@Aeyk  can you please take a look?

Issue #183, opened 2024-03-29
need argument "--proxy-headers" if keycloak is used behind a proxy


roumano:
##### SUMMARY

If keycloak is behind a proxy, in some case, we need to add the argument `--proxy-headers` of the start of the service

See [Configure the reverse proxy headers](https://www.keycloak.org/server/reverseproxy) for the detail in keycloak configuration

i propose this change on `templates/keycloak.service.j2` :

```
{% if keycloak_quarkus_start_dev %}
ExecStart={{ keycloak.home }}/bin/kc.sh start-dev{%if keycloak_quarkus_proxy_headers is defined %} --proxy-headers {{keycloak_quarkus_proxy_headers}}{% endif %}{{ '\n' }}
{% else %}
ExecStart={{ keycloak.home }}/bin/kc.sh start --optimized{%if keycloak_quarkus_proxy_headers is defined %} --proxy-headers {{keycloak_quarkus_proxy_headers}}{% endif %}{{ '\n' }}
{% endif %}
```

instead of the actual : 
```
{% if keycloak_quarkus_start_dev %}
ExecStart={{ keycloak.home }}/bin/kc.sh start-dev
{% else %}
ExecStart={{ keycloak.home }}/bin/kc.sh start --optimized
{% endif %}

```

So, it's can be used with this variable : `keycloak_quarkus_proxy_headers: xforwarded`

##### ISSUE TYPE
 - Feature Idea

PR #187, opened 2024-04-15
Permit parse reverse proxy headers

roumano:
- Via created a new optional variable : keycloak_quarkus_proxy_headers
- Fix enhancement #183
- see https://www.keycloak.org/server/reverseproxy about the official keycloak documentation

### DIFF
--- a/roles/keycloak_quarkus/README.md
+++ b/roles/keycloak_quarkus/README.md
@@ -54,7 +54,7 @@ Role Defaults
 |`keycloak_quarkus_https_trust_store_enabled`| Enalbe confiugration of a trust store | `False` |
 |`keycloak_quarkus_trust_store_file`| The file pat to the trust store | `{{ keycloak.home }}/conf/trust_store.p12` |
 |`keycloak_quarkus_trust_store_password`| Password for the trust store | `""` |
-
+|`keycloak_quarkus_proxy_headers`| Parse reverse proxy headers (`forwarded` or `xforwardedPassword`) | `""` |
 
 * Hostname configuration
 
--- a/roles/keycloak_quarkus/templates/keycloak.service.j2
+++ b/roles/keycloak_quarkus/templates/keycloak.service.j2
@@ -8,10 +8,10 @@ Type=simple
 EnvironmentFile=-{{ keycloak_quarkus_sysconf_file }}
 PIDFile={{ keycloak_quarkus_service_pidfile }}
 {% if keycloak_quarkus_start_dev %}
-ExecStart={{ keycloak.home }}/bin/kc.sh start-dev
-{% else %}
-ExecStart={{ keycloak.home }}/bin/kc.sh start --optimized
-{% endif %}
+ExecStart={{ keycloak.home }}/bin/kc.sh start-dev{% if keycloak_quarkus_proxy_headers is defined %} --proxy-headers {{ keycloak_quarkus_proxy_headers }}{% endif -%}{{ '\n' }}
+{% else -%}
+ExecStart={{ keycloak.home }}/bin/kc.sh start --optimized{% if keycloak_quarkus_proxy_headers is defined %} --proxy-headers {{ keycloak_quarkus_proxy_headers }}{% endif -%}{{ '\n' }}
+{%- endif %}
 User={{ keycloak.service_user }}
 Group={{ keycloak.service_group }}
 {% if keycloak_quarkus_service_restart_always %}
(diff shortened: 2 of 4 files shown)


roles/keycloak_quarkus/README.md
```diff
@@ -54,7 +54,7 @@ Role Defaults
 |`keycloak_quarkus_https_trust_store_enabled`| Enalbe confiugration of a trust store | `False` |
 |`keycloak_quarkus_trust_store_file`| The file pat to the trust store | `{{ keycloak.home }}/conf/trust_store.p12` |
 |`keycloak_quarkus_trust_store_password`| Password for the trust store | `""` |
-
+|`keycloak_quarkus_proxy_headers`| Parse reverse proxy headers (`forwarded` or `xforwardedPassword`) | `""` |
```
guidograzioli, line 57: This needs to be spec-ed in meta/argument_specs.yml; in the description please mention it overrides the deprecated `keycloak_quarkus_proxy_mode` argument

roles/keycloak_quarkus/templates/keycloak.service.j2
```diff
@@ -8,10 +8,10 @@ Type=simple
 EnvironmentFile=-{{ keycloak_quarkus_sysconf_file }}
 PIDFile={{ keycloak_quarkus_service_pidfile }}
 {% if keycloak_quarkus_start_dev %}
-ExecStart={{ keycloak.home }}/bin/kc.sh start-dev
-{% else %}
-ExecStart={{ keycloak.home }}/bin/kc.sh start --optimized
-{% endif %}
+ExecStart={{ keycloak.home }}/bin/kc.sh start-dev{% if keycloak_quarkus_proxy_headers is defined %} --proxy-headers {{ keycloak_quarkus_proxy_headers }}{% endif -%}{{ '\n' }}
```
guidograzioli, line 11: Not here please, as we try to avoid cmd-line arguments; put an if block inside the config template instead.

roumano: Do you prefere double if linked (personally i  think it's not easy to read ) : 
```
{% if keycloak_quarkus_start_dev %}
{% if keycloak_quarkus_proxy_headers is defined %}
ExecStart={{ keycloak.home }}/bin/kc.sh start-dev --proxy-headers {{ keycloak_quarkus_proxy_headers }}
{% else -%}
ExecStart={{ keycloak.home }}/bin/kc.sh start-dev
{%- endif %}
{% else -%}
{% if keycloak_quarkus_proxy_headers is defined %}
ExecStart={{ keycloak.home }}/bin/kc.sh start --optimized --proxy-headers {{ keycloak_quarkus_proxy_headers }}
{% else -%}
ExecStart={{ keycloak.home }}/bin/kc.sh start --optimized
{% endif %}
{% endif %}
```

Or a only if (but more complex and elif condition : 
```
{% if keycloak_quarkus_start_dev and keycloak_quarkus_proxy_headers is defined %}
ExecStart={{ keycloak.home }}/bin/kc.sh start-dev --proxy-headers {{ keycloak_quarkus_proxy_headers }}
{% elif keycloak_quarkus_start_dev and keycloak_quarkus_proxy_headers is not defined %}
ExecStart={{ keycloak.home }}/bin/kc.sh start-dev
{% elif not keycloak_quarkus_start_dev and keycloak_quarkus_proxy_headers is defined %}
ExecStart={{ keycloak.home }}/bin/kc.sh start --optimized --proxy-headers {{ keycloak_quarkus_proxy_headers }}
{% else -%}
ExecStart={{ keycloak.home }}/bin/kc.sh start --optimized
{% endif %}
```

roumano: I've updated my MR with the second choix, but i can use the first one if you prefer

guidograzioli: Oh sorry, I was unclear; I meant, no arguments added to `kc.sh`; arguments should go instead in the config file template for [keycloak.conf ](https://github.com/ansible-middleware/keycloak/blob/main/roles/keycloak_quarkus/templates/keycloak.conf.j2)

roumano: I confirm it's much better with this solution.
As i follow "stupidly" the official documentation, i was thinking it's not possible to have it on the keycloak.conf.
i confirm, it's working as expected with configuration on `keycloak.conf`.

guidograzioli (approved):
LGTM, thanks!

roumano:
you welcome.

It's possible to release a new version (lastest is 2.1.0 done the 28 february), so it's can easily deploy on production server ?

Issue #188, opened 2024-04-15
keycloak_quarkus:  allow setting "sensitive options" using a Java KeyStore file


hwo-wd:
##### SUMMARY
As per <https://www.keycloak.org/server/configuration#_setting_sensitive_options_using_a_java_keystore_file> the idea is to set these [three configuration properties/env variables](https://www.keycloak.org/server/all-config?q=config-keystore):
* `KC_CONFIG_KEYSTORE`
* `KC_CONFIG_KEYSTORE_PASSWORD`
* `KC_CONFIG_KEYSTORE_TYPE`

The idea is to set sensitive fields[^1] in the key store instead; for the time being, the PCI-DSS4 auditor wants to have
* `keycloak_quarkus_db_pass`

in the keystore, to "provide an additional layer of obstruction"...

[^1]: these need to be in `keycloak.conf` though, since Quarkus doesn't seem to provide a similar option as of now.

Note that this item is different to #172 as this one is about the configuration options, while the former is about a vault provider for client secrets etc.

##### ISSUE TYPE
 - Feature Idea

PR #189, opened 2024-04-16
#188: add support for configuration key store

hwo-wd:
`keytool` might not be the best tool for this, but I didn't find any better: `openssl` does support PKCS12 stores, but does not allow for creating PBEs (password-based entries); 

Note that this has a breaking change since variables had to be renamed to reduce confusion.

Open for any improvements, but I didn't find a more ansiblish way to do it since keytool doesn't allow for reading the actual secret values, only whether it exists or not.

Moreover let me know if you see more variables from `keycloak.conf` that would fit into this key store, thanks.

### DIFF
--- a/roles/keycloak_quarkus/tasks/config_store.yml
+++ b/roles/keycloak_quarkus/tasks/config_store.yml
@@ -1,16 +1,4 @@
 ---
-- name: "Check if keytool exists in path"
-  block:
-    - name: "Attempt to run keytool"
-      ansible.builtin.command: keytool -help
-      register: keytool_check
-      ignore_errors: true
-
-    - name: "Fail when no keytool found"
-      when: keytool_check.rc != 0
-      ansible.builtin.fail:
-        msg: "keytool NOT found in the PATH, but is required for setting up the configuration key store"
-
 - name: "Initialize configuration key store variables to be written"
   ansible.builtin.set_fact:
     store_items:
--- a/roles/keycloak_quarkus/tasks/config_store.yml
+++ b/roles/keycloak_quarkus/tasks/config_store.yml
@@ -1,0 +1,64 @@
+---
+- name: "Check if keytool exists in path"
+  block:
+    - name: "Attempt to run keytool"
+      ansible.builtin.command: keytool -help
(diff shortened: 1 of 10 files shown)


guidograzioli:
I am thinking writing a general keytool module for the middleware common collection, but for the time being this is cool.
But, I'd keep the two "old" renamed vars, setting them to the new parameters value. We can mark them deprecated and remove in the next release, without breaking (so basically, temporarily, keycloak_quarkus_key_store_file default to keycloak_quarkus_https_key_store_file), wdyt?

roles/keycloak_quarkus/tasks/config_store.yml
```diff
@@ -0,0 +1,64 @@
+---
+- name: "Check if keytool exists in path"
+  block:
```
guidograzioli, line 3: This block happen quite late in the deployment; would you mind moving it to the prereq tasks? change_when below can safely be set to false

hwo-wd: The check now happens after having installed JVM in `fastpackages.yml`, so keytool should definitely be there, but  a check never hurts either ;)

hwo-wd:
> I am thinking writing a general keytool module for the middleware common collection, but for the time being this is cool. But, I'd keep the two "old" renamed vars, setting them to the new parameters value. We can mark them deprecated and remove in the next release, without breaking (so basically, temporarily, keycloak_quarkus_key_store_file default to keycloak_quarkus_https_key_store_file), wdyt?

I've added this, probably a bit too compliced but it might serve as a base for other roles.

guidograzioli:
new linter rules on ahub driving me crazy, please rebase on main (again); sorry bout that

hwo-wd:
> new linter rules on ahub driving me crazy, please rebase on main (again); sorry bout that

no worries -> done.

guidograzioli:
OK; i'll release .2 as final 2.1 (with all the linter ****); then I'll merge and we can eventually start 2.2
Thanks for your contribution!

guidograzioli (approved):
LGTM

PR #195, opened 2024-04-17
Added hostname-strict-https option

### LINKED ISSUE #193: Unable to set hostname-strict-https=false
##### SUMMARY

I am try to deploy middleware_automation.keycloak.keycloak_quarkus with the following playbook:
```yaml
---
- name: Playbook for Keycloak Hosts
  hosts: all
  vars:
    keycloak_quarkus_proxy_mode: edge
    keycloak_quarkus_http_enabled: true
    keycloak_quarkus_hostname_strict: false
   keycloak_quarkus_hostname_strict_backchannel: false
    
    keycloak_quarkus_admin_pass: "remembertochangeme"
    keycloak_quarkus_configure_firewalld: true
  roles:
    - middleware_automation.keycloak.keycloak_quarkus
```

But when I try to connect using HTTP, its admin page keeps blank with a loading animation. To make it work I have to manually put `hostname-strict-https=false` and restart the keycloak service.

Therefore, I am requesting to add `hostname-strict-https` to the role variables.

##### ISSUE TYPE
 - Add role variable `hostname-strict-https = false | true [default]`

InfoSec812:
Resolves #193 

Added a new option `keycloak_quarkus_hostname_strict_https` with no default value.

If the option is set, the Jinja template for `keycloak.conf.j2` checks to see if the value is either `true` or `false` and applies it appropriately

### DIFF
--- a/roles/keycloak_quarkus/templates/keycloak.conf.j2
+++ b/roles/keycloak_quarkus/templates/keycloak.conf.j2
@@ -10,10 +10,10 @@ db-password={{ keycloak_quarkus_db_pass }}
 {% endif %}
 {% endif %}
 
-{% if keycloak_quarkus_hostname_strict_https and keycloak_quarkus_hostname_strict_https is sameas true -%}
+{% if keycloak_quarkus_hostname_strict_https is defined and keycloak_quarkus_hostname_strict_https is sameas true -%}
 hostname-strict-https=true
 {% endif -%}
-{% if keycloak_quarkus_hostname_strict_https and keycloak_quarkus_hostname_strict_https is sameas false -%}
+{% if keycloak_quarkus_hostname_strict_https is defined and keycloak_quarkus_hostname_strict_https is sameas false -%}
 hostname-strict-https=false
 {% endif -%}
 
--- a/roles/keycloak_quarkus/templates/keycloak.conf.j2
+++ b/roles/keycloak_quarkus/templates/keycloak.conf.j2
@@ -10,6 +10,10 @@ db-password={{ keycloak_quarkus_db_pass }}
 {% endif %}
 {% endif %}
 
+{% if keycloak_quarkus_hostname_strict_https -%}
+hostname-strict-https={{ keycloak_quarkus_hostname_strict_https }}
+{% endif -%}
+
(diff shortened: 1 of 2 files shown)


roles/keycloak_quarkus/templates/keycloak.conf.j2
```diff
@@ -10,6 +10,10 @@ db-password={{ keycloak_quarkus_db_pass }}
 {% endif %}
 {% endif %}
 
+{% if keycloak_quarkus_hostname_strict_https -%}
+hostname-strict-https={{ keycloak_quarkus_hostname_strict_https }}
```
guidograzioli, line 14: you need `| lower` or python will emit False instead of false, which quarkus doesn't like

guidograzioli: plus, please add the role parameter to meta/argument_specs.yml, with `required: no` and telling in the description that it only applies when `keycloak_quarkus_frontend_url` is not set

guidograzioli (changes requested):
just a minor change needed, thanks for the contribution

roles/keycloak_quarkus/templates/keycloak.conf.j2
```diff
@@ -10,6 +10,10 @@ db-password={{ keycloak_quarkus_db_pass }}
 {% endif %}
 {% endif %}
 
+{% if keycloak_quarkus_hostname_strict_https -%}
```
guidograzioli, line 13: `if keycloak_quarkus_hostname_strict_https is defined`

InfoSec812:
@guidograzioli I modified it to better handle ensuring that the value is a boolean and it should address your concern as well.

guidograzioli:
Thanks for updating, but the specification for the new parameter is still needed here:
https://github.com/ansible-middleware/keycloak/blob/main/roles/keycloak_quarkus/meta/argument_specs.yml#L335

InfoSec812:
@guidograzioli OK, I think that latest change should cover it.

Thanks!

guidograzioli (approved):
LGTM, thanks

PR #196, opened 2024-04-18
Keystore based vault SPI

### LINKED ISSUE #172: Vault support
##### SUMMARY

Keycloak has support for a [Vault SPI](https://www.keycloak.org/server/vault). Two providers are available by default in Keycloak, including a file and JKS provider.

These have to be added at build time, with the `--vault=file` or `--vault=keystore` args.

It would be nice if this module supported said functionality.

##### ISSUE TYPE
 - Feature Idea

guidograzioli:
Add parameters that allow to configure keystore-based vault SPI:

Role Defaults
-------------
|`keycloak_quarkus_ks_vault_enabled`| Whether to enable the vault SPI | `false` |
|`keycloak_quarkus_ks_vault_file`| The keystore path for the vault SPI | `{{ keycloak_quarkus_config_dir }}/keystore.p12` |
|`keycloak_quarkus_ks_vault_type`| Type of the keystore used for the vault SPI | `PKCS12` |

Role Variables
--------------

| Variable | Description | Required |
|:---------|:------------|----------|
|`keycloak_quarkus_ks_vault_pass`| The password for accessing the keystore vault SPI | `no` |

See: https://www.keycloak.org/server/vault#_configuring_the_java_keystore_based_vault
Also see molecule/quarkus/converge.yml for usage

Fix #172

### DIFF
--- a/molecule/quarkus/converge.yml
+++ b/molecule/quarkus/converge.yml
@@ -1,16 +1,20 @@
 ---
 - name: Converge
   hosts: all
-  vars: 
+  vars:
     keycloak_quarkus_admin_pass: "remembertochangeme"
     keycloak_admin_password: "remembertochangeme"
     keycloak_realm: TestRealm
     keycloak_quarkus_host: instance
     keycloak_quarkus_log: file
-    keycloak_quarkus_https_key_file_enabled: True
+    keycloak_quarkus_log_level: debug
+    keycloak_quarkus_https_key_file_enabled: true
     keycloak_quarkus_key_file: "/opt/keycloak/certs/key.pem"
     keycloak_quarkus_cert_file: "/opt/keycloak/certs/cert.pem"
     keycloak_quarkus_log_target: /tmp/keycloak
+    keycloak_quarkus_ks_vault_enabled: true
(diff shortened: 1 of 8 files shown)


hwo-wd:
@guidograzioli acc. to the [docs](https://www.keycloak.org/server/vault#_enabling_a_vault), I guess you'd also need to notify and adapt `rebuild_config.yml` accordingly. In my experience, not rebuilding the configuration does the trick just as well, but sometimes (looking at you, ispn cache) it behaves weird. So best to rebuild and be safe.

hwo-wd:
> > @guidograzioli acc. to the [docs](https://www.keycloak.org/server/vault#_enabling_a_vault), I guess you'd also need to notify and adapt `rebuild_config.yml` accordingly. In my experience, not rebuilding the configuration does the trick just as well, but sometimes (looking at you, ispn cache) it behaves weird. So best to rebuild and be safe.
> 
> Any keycloak.conf change will trigger the rebuild afaics (https://github.com/ansible-middleware/keycloak/actions/runs/8735278501/job/23967743420#step:5:333) using the updated properties. Do you mean it is not enough?

You're right, I forgot/didn't check my own code, sry... 

image

guidograzioli:
lmao, merging, thx

Issue #190, opened 2024-04-16
Discussion: possibility to "get rid of" `KEYCLOAK_ADMIN` and `KEYCLOAK_ADMIN_PASSWORD` env vars once the stage was bootstrapped


hwo-wd:
##### SUMMARY
These two env vars allow to set the initial admin and kc takes care of creating this initial user if it doesn't exit yet. With other words, these are only really required on a new installation but from a security perspective the do not necessarily have to reside in `/etc/sysconfig/keycloak` where the whole world (with access to the machine) can read these and thus basically have a root-login.

I've been thinking whether there's a possibility to reduce the exposure of this sensitive data in an automated way, something like keeping them until KC has fully started, then removing them, but I do not really like any of my ideas so far.
So this issue is really more a brainstorming on how one could approach this as part of this role.

(Untested, but should be doable) manual workaround by the operator would be:
* to manually change the `KEYCLOAK_ADMIN_PASSWORD` to a dummy value once the environment has been fully initialized (NB: the role doesn't support an empty password as of now)
* to introduce a flag whether these env variables should be set, but this boils down to the former option pretty much

Note: reading the KC source, these 2 can only be set as environment variables, i.e., there is no option to move these into a trust store which could only be read by `root` and `keycloak`...

##### ISSUE TYPE
 -  Question/Discussion

guidograzioli:
oh wow this being 644 is huge issue: https://github.com/ansible-middleware/keycloak/blob/main/roles/keycloak_quarkus/tasks/systemd.yml#L9, definitely needs to be 400
well spotted

As for the envvars I am aware of their single-shot usage; and have been reluctant to do anything without proper thinking, so anybody with idea pls come forward :)

guidograzioli:
That said, if somebody has enough privileges on the machine to read /etc/sysconfig/ , no envvar protection can really be implemented, because he would be able to read /proc/

PR #197, opened 2024-04-19
#190: remove `keycloak_quarkus_admin_user[_pass]` once keycloak is bootstrapped

hwo-wd:
@guidograzioli this is my take on #190:
This approach could be troublesome when the state file (`/etc/sysconfig/keycloak`) contains the bootstrapped mnemonicer already, but we are switching DBs where the new DB is completely empty.
But I guess this edge case would be ok favoring hiding these sensitive credentials on the long term, worst case one needs to set `keycloak_quarkus_purge_admin_credentials_after_bootstrapping=True` and then run the playbook once more.

Wdyt?

### DIFF
--- a/roles/keycloak_quarkus/README.md
+++ b/roles/keycloak_quarkus/README.md
@@ -152,6 +152,14 @@ Role Variables
 |`keycloak_quarkus_admin_url`| Base URL for accessing the administration console, including scheme, host, port and path | `no` |
 |`keycloak_quarkus_ks_vault_pass`| The password for accessing the keystore vault SPI | `no` |
 
+Role custom facts
+-----------------
+
+The role uses the following [custom facts](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_vars_facts.html#adding-custom-facts) found in `/etc/ansible/facts.d/keycloak.fact` (and thus identified by the `ansible_local.keycloak.` prefix):
+
+| Variable | Description |
+|:---------|:------------|
+|`general.bootstrapped` | A custom fact indicating whether this role has been used for bootstrapping keycloak on the respective host before; set to `false` (e.g., when starting off with a new, empty database) ensures that the initial admin user as defined by `keycloak_quarkus_admin_user[_pass]` gets created |
 
 License
 -------
--- a/roles/keycloak_quarkus/tasks/main.yml
+++ b/roles/keycloak_quarkus/tasks/main.yml
@@ -96,11 +96,6 @@
 - name: "Start and wait for keycloak service"
   ansible.builtin.include_tasks: start.yml
 
-- name: Check service status
-  ansible.builtin.command: "systemctl status keycloak"
-  register: keycloak_service_status
-  changed_when: false
-
 - name: Link default logs directory
   ansible.builtin.file:
     state: link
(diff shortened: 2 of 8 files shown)


guidograzioli:
I like the general mechanic of the changeset; let me review it thru

roles/keycloak_quarkus/templates/keycloak-sysconfig.j2
```diff
@@ -1,6 +1,10 @@
 {{ ansible_managed | comment }}
+{% if not keycloak_quarkus_purge_admin_credentials_after_bootstrapping or not keycloak_quarkus_internal_bootstrapped %}
 KEYCLOAK_ADMIN={{ keycloak_quarkus_admin_user }}
 KEYCLOAK_ADMIN_PASSWORD='{{ keycloak_quarkus_admin_pass }}'
+{% else %}
+{{ keycloak.bootstrap_mnemonic }}
```
guidograzioli, line 6: Let's always add this line no matter what; it can turn useful in other scenarios.

roles/keycloak_quarkus/tasks/main.yml
```diff
@@ -96,15 +96,30 @@
...
+    name: "{{ keycloak.service_name }}"
+  register: keycloak_service_status
+  changed_when: false
+
+- name: "Trigger bootstrapped notification: remove `keycloak_quarkus_admin_user[_pass]` env vars"
+  when:
+    - keycloak_quarkus_purge_admin_credentials_after_bootstrapping
+    - not keycloak_quarkus_internal_bootstrapped             # it was not bootstrapped prior to the current role's execution
+    - keycloak_service_status.status.ActiveState == "active" # but it is now
+  ansible.builtin.set_fact:
```
guidograzioli, line 118: In addition to this (and the mnemonic added to the sysconfig file), I propose to also create a custom fact https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_vars_facts.html#facts-d-or-local-facts

With that, we can get rid of the grep command (the custom fact will just be there); and we also get rid of the `keycloak_quarkus_purge_admin_credentials_after_bootstrapping` parameter altogether. We make the `true` the default behaviour; and we tell to delete the custom fact in case credentials are needed for any reason

guidograzioli (commented):
I left my comments; the general gist is that I'd prefer not to have a default `true` parameter, but just do your implemented behaviour by default. Using a custom fact (being bootstrapped or not), we can tell users to delete it on the target node to execute the `false` behaviour (adding the credentials to sysconfig).

_mind_ I am not requesting changes, just discussing!

roles/keycloak_quarkus/tasks/main.yml
hwo-wd: didn't know about these local/static facts, but they look nice and serve our use case well.
Not sure about keeping `{{ keycloak.bootstrap_mnemonic }}` then, but I'm not religious about it; I'll start with the local facts and you let me know whether to keep the mnemonicer

guidograzioli: please keep it!

hwo-wd:
pushed an update with custom ansible facts; @guidograzioli please review it once more; I'd squash the commits then

roles/keycloak_quarkus/README.md
```diff
@@ -142,6 +143,15 @@ Role Variables
 |`keycloak_quarkus_admin_url`| Base URL for accessing the administration console, including scheme, host, port and path | `no` |
 
 
+Role custom facts
```
guidograzioli, line 146: :+1: :+1:

guidograzioli (approved):
LGTM

hwo-wd:
> LGTM

thx; linter issues resolved too -> all done

guidograzioli:
Nice work!

PR #199, opened 2024-04-23
Add wait_for systemd logic

guidograzioli:


roles/keycloak_quarkus/tasks/restart.yml
```diff
@@ -1,5 +1,6 @@
 ---
 - name: "Restart and enable {{ keycloak.service_name }} service"
+  throttle: 1
```
rpelisse, line 3: Should we consider adding this to other use of systemd (in other collections)?

roles/keycloak_quarkus/templates/keycloak.service.j2
```diff
@@ -4,25 +4,30 @@ Description=Keycloak Server
...
 {% endif %}
 RestartSec={{ keycloak_quarkus_service_restartsec }}
-{% if keycloak_quarkus_http_port|int < 1024 or keycloak_quarkus_https_port|int < 1024 %}
+{% if keycloak_quarkus_http_port | int < 1024 or keycloak_quarkus_https_port | int < 1024 %}
 AmbientCapabilities=CAP_NET_BIND_SERVICE
 {% endif %}
+{% if keycloak_quarkus_systemd_wait_for_port %}
+ExecStartPost=/usr/bin/timeout {{ keycloak_quarkus_systemd_wait_for_timeout }} sh -c 'while ! ss -H -t -l -n sport = :{{ keycloak_quarkus_https_port }} | grep -q "^LISTEN.*:{{ keycloak_quarkus_https_port }}"; do sleep 1; done && /bin/sleep {{ keycloak_quarkus_systemd_wait_for_delay }}'
+{% endif %}
+{% if keycloak_quarkus_systemd_wait_for_log %}
```
rpelisse, line 28: Maybe I misreading this, but why is there two if with the exact same clause? Both are ` if keycloak_quarkus_systemd_wait_for_log`, so why not regroup? Or maybe one of them is wrong?

rpelisse (changes requested):
Misc comments, otherwise looks good to me.

roles/keycloak_quarkus/handlers/main.yml
```diff
@@ -9,9 +9,9 @@
 - name: "Restart {{ keycloak.service_name }}"
   ansible.builtin.include_tasks: restart.yml
   listen: "restart keycloak"
-
 - name: "Print deprecation warning"
   ansible.builtin.fail:
     msg: "Deprecation warning: you are using the deprecated variable '{{ deprecated_variable | d('NotSet') }}', check docs on how to upgrade."
   ignore_errors: true
+  failed_when: false
```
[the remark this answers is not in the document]
guidograzioli: it's matter of colors.... fail shows in red (but then it's ignored); debug shows in green, unless we have changed_when: always in which case it turns orange (we have orange in jws)

roles/keycloak_quarkus/tasks/restart.yml
guidograzioli: yep

roles/keycloak_quarkus/templates/keycloak.service.j2
guidograzioli: they differ in the rightmost word.... long names indeed are painful on the eyes

rpelisse (approved):

PR #201, opened 2024-04-30
Providers config and custom providers

### LINKED ISSUE #131: Providers add to script
Hi Team, 

thank you for great ansible for keycloak. I wonder if you are planing to add that would b possible during installation add providers or themes to keycloak?
I'm interested in providers because I'm deploying keycloak cluster to AWS and there I have to install few providers and also change HA settings. 
And also we are using PrivacyIDEA keycloak plugin. 
Another think is we are using mysql would be possible to add not that is hard to add but it is handy if would be there already.

I modify my script and add extra task for providers

- name: Include binaries for providers
  when: keycloak_quarkus_providers is defined
  ansible.builtin.include_tasks: providers.yml
  register: keycloak_provides
  tags:
    - providers

and task to for build 

- name: Building Keycloak 
  when: keycloak_provides is defined
  ansible.builtin.command: "{{ keycloak.home }}/bin/kc.sh build"
  changed_when: False
  become: yes
  notify:
    - restart keycloak

providers.yml
---
- name: Download JAR files for providers
  get_url:
    url: "{{ item }}"
    dest: "{{ keycloak_quarkus_providers_dir }}{{ item | regex_replace('.+/(\\w+-\\d+\\.\\d+\\.\\d+\\.jar)', '\\1') }}"
  loop: "{{ keycloak_quarkus_providers }}"

And for AWS deployment I modify cache-ispn.xml. 

I did very simple but probably you can do better than me, because my knowledge is very basic for ansible and keycloak.
Hope not asking to much. thanks and keep doing good work.
Ales

guidograzioli:
#### Configuring providers

|`keycloak_quarkus_providers`| List of provider definitions; see below | `[]` |

Provider definition:

```yaml
keycloak_quarkus_providers:
  - id: http-client                         # required
    spi: connections                        # required if url is not specified
    default: true                           # optional, whether to set default for spi, default false
    restart: true                           # optional, whether to restart, default true
    url: https://.../.../custom_spi.jar     # optional, url for download
    properties:                             # optional, list of key-values
      - key: default-connection-pool-size
        value: 10
```

the definition above will generate the same configuration as the following build command:

```
bin/kc.sh build --spi-connections-provider=http-client --spi-connections-http-client-default-connection-pool-size=10
```

Fix #131

### DIFF
--- a/molecule/quarkus/converge.yml
+++ b/molecule/quarkus/converge.yml
@@ -19,6 +19,16 @@
     keycloak_quarkus_systemd_wait_for_timeout: 20
     keycloak_quarkus_systemd_wait_for_delay: 2
     keycloak_quarkus_systemd_wait_for_log: true
+    keycloak_quarkus_providers:
+      - id: http-client
+        spi: connections
+        default: true
+        restart: true
+        properties:
+          - key: default-connection-pool-size
+            value: 10
+      - id: spid-saml
+        url: https://github.com/italia/spid-keycloak-provider/releases/download/24.0.2/spid-provider.jar
   roles:
     - role: keycloak_quarkus
     - role: keycloak_realm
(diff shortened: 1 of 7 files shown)


rpelisse (approved):
LGTM

PR #202, opened 2024-04-30
Customize jdbc driver downloads, optional authentication

### LINKED ISSUE #200: Allow for downloading JDBC drivers from alternate URLs and allow for using username/password
##### SUMMARY
In some environments, access to Maven Central is not allowed, but those environments often have a local repository manager which "proxies" maven central. Sometimes those repository managers require authentication to download. It would be useful to be able to:
1. Set an alternative URL for downloading JDBC drivers
2. Set credentials for authentication to download said JDBC drivers.

##### ISSUE TYPE
 - Feature Idea

InfoSec812:
Add three new parameters which allow for overriding the download URL for the JDBC driver and also specifying credentials for downloading those JDBC drivers.

| Variable | Description | Required |
|:---------|:------------|:---------|
|`keycloak_quarkus_jdbc_download_url`| Override the default Maven Central download URL for the JDBC driver | `no` |
|`keycloak_quarkus_jdbc_download_user`| Set a username with which to authenticate when downloading JDBC drivers from an alternative location | `no` |
|`keycloak_quarkus_jdbc_download_pass`| Set a password with which to authenticate when downloading JDBC drivers from an alternative location (requires keycloak_quarkus_jdbc_download_user) | `no` |

Resolves #200

### DIFF
--- a/roles/keycloak_quarkus/meta/argument_specs.yml
+++ b/roles/keycloak_quarkus/meta/argument_specs.yml
@@ -372,6 +372,15 @@ argument_specs:
                 description: "Activation delay for service systemd unit (seconds)"
                 default: 10
                 type: 'int'
+            keycloak_quarkus_jdbc_download_url:
+                description: "Override the default Maven Central download URL for the JDBC driver"
+                type: "str"
+            keycloak_quarkus_jdbc_download_user:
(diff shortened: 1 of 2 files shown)


guidograzioli (approved):
LGTM; thanks for the contribution!

Issue #208, opened 2024-05-03
Copying Key Material


Footur:
##### SUMMARY
The default directory for certificates in `keycloak_quarkus` is `"{{ keycloak_home }}/certs"`. This directory is not created by the role. Also, I am missing a task that copies the key material to `"{{ keycloak_home }}/certs"`. This should be done before the first start of Keycloak.

##### ISSUE TYPE
 - Feature Idea

guidograzioli:
Correct, at the moment the role expects the certificates to be already in place (ideally, one would install them under /etc/pki/tls/ ). We could have a few tasks that manage to use local files or download them and copy them to target nodes.

Footur:
@guidograzioli What do you think about using the directories
- `/etc/pki/tls/private` for the private key and
- `/etc/pki/tls/cert` for the certificate
as default values?

guidograzioli:
Since what you're proposing is the default in Red Hat distributions, I without any doubt second that!

Footur:
@guidograzioli Can you create a new release please? :pray:

PR #210, opened 2024-05-03
Enable copying of key material

Footur:
This pull request updates the configuration to use the standard Red Hat Enterprise Linux (RHEL) default path for TLS certificates.

Also, it copies the private key and certificate to the target host.

 #208


guidograzioli (changes requested):
thanks for the PR; the change looks good, but can i ask to also update the quarkus molecule test to use the functionality? (instead of using the prepare.yml to copy the files)

Footur:
@guidograzioli I have updated the Molecule test. Does it meet your expectations?

guidograzioli (approved):

PR #223, opened 2024-05-13
Allow for custom providers hosted on maven repositories

### LINKED ISSUE #222: keycloak_quarkus: Add support for custom providers hosted on MVN
Extend the way we can [configure providers](https://github.com/ansible-middleware/keycloak/tree/main/roles/keycloak_quarkus#configuring-providers) such that mvn SPIs are supported as well; this is especially helpful when hosting things on MVN Central or on a Github MVN repository. For the latter the usage of a username and a PAT is required.

hwo-wd:
Extends support for custom providers ("SPIs"): while previously only http downloads were supported, the new provider definition allows for maven downloads (e.g., Apache Maven Central,  Github Apache Maven (requires authn via a PAT)).

The new providers definition looks like:

```yaml
keycloak_quarkus_providers:
  - id: http-client                         # required
    spi: connections                        # required if url is not specified
  - id: http-client                         # required; "{{ id }}.jar" identifies the file name on RHBK
    spi: connections                        # required if neither url, local_path nor maven are specified; required for setting properties
    default: true                           # optional, whether to set default for spi, default false
    restart: true                           # optional, whether to restart, default true
    url: https://.../.../custom_spi.jar     # optional, url for download
    url: https://.../.../custom_spi.jar     # optional, url for download via http
    local_path: my_theme_spi.jar            # optional, path on local controller for SPI to be uploaded
    maven:                                  # optional, for download using maven
      repository_url: https://maven.pkg.github.com/OWNER/REPOSITORY # optional, maven repo url
      group_id:  my.group                   # optional, maven group id
      artifact_id: artifact                 # optional, maven artifact id
      version: 24.0.4                       # optional, defaults to latest
      username:  user                       # optional, cf. https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-apache-maven-registry#authenticating-to-github-packages
      password: pat                         # optional, provide a PAT for accessing Github's Apache Maven registry
    properties:                             # optional, list of key-values
```

* `lxml` is a new python dependency for this collection (however, it was already present because it is also a dependency of `middleware_automation.common`

Close #222

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -1,5 +1,5 @@
 ---
 collections:
   - name: middleware_automation.common
+    version: ">=1.2.1"
   - name: ansible.posix
-  - name: community.general # for `maven_artifact`
--- a/requirements.yml
+++ b/requirements.yml
@@ -2,3 +2,4 @@
 collections:
   - name: middleware_automation.common
   - name: ansible.posix
+  - name: community.general # for `maven_artifact`
--- a/molecule/quarkus/converge.yml
+++ b/molecule/quarkus/converge.yml
@@ -39,6 +39,8 @@
           version: 24.0.4 # optional
           # username: myUser # optional
           # password: myPAT # optional
+      # - id: my-static-theme
+      #   local_path: /tmp/my-static-theme.jar
     keycloak_quarkus_policies:
       - name: "xato-net-10-million-passwords.txt"
         url: "https://github.com/danielmiessler/SecLists/raw/master/Passwords/xato-net-10-million-passwords.txt"
--- a/molecule/quarkus/converge.yml
+++ b/molecule/quarkus/converge.yml
@@ -31,6 +31,14 @@
             value: 10
       - id: spid-saml
         url: https://github.com/italia/spid-keycloak-provider/releases/download/24.0.2/spid-provider.jar
+      - id: keycloak-kerberos-federation
+        maven:
+          repository_url: https://repo1.maven.org/maven2/ # https://mvnrepository.com/artifact/org.keycloak/keycloak-kerberos-federation/24.0.4
+          group_id: org.keycloak
+          artifact_id: keycloak-kerberos-federation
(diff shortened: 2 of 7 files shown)


requirements.yml
```diff
@@ -2,3 +2,4 @@
 collections:
   - name: middleware_automation.common
   - name: ansible.posix
+  - name: community.general # for `maven_artifact`
```
guidograzioli, line 5: ouch; that's a big no-no for the downstream version of the collection

hwo-wd:
WIP: I'll piggy-pack local file support right into this PR: this way we can support themes just like any other SPI.

requirements.yml
hwo-wd: hmm, any other options that come to your mind? At the end of the day it's just a plain http download w/ authentication, but re-inventing the wheel wrt. authn against github repo seems more than smelly...

guidograzioli: I am considering forking the module inside middleware_automation.common (redhat.runtimes_common), so it can go supported. It would also lower the changeset bar down from breaking_change to minor_change; please continue your work, I'll find a solution

hwo-wd: Please also note the addition of `lxml` in `requirements.txt`, so I guess this is breaking for users already making use of providers on way or the other...

guidograzioli: We already have lxml in the common dependency reqs :); but i'll remove the boto dependency for s3 in the forked module. I'd say it marks it has a major change (user may have to make changes if they were already installing providers by themselves)

guidograzioli: You can already do the change in this PR please:
community.general.maven_artifact -> middleware_automation.common.maven_artifact

guidograzioli: https://github.com/ansible-middleware/common/pull/24

hwo-wd: All done.

guidograzioli (approved):
all good

Issue #224, opened 2024-05-14
keycloak_quarkus: Add support for policy files (e.g. password blacklists)


hwo-wd:
<https://access.redhat.com/documentation/en-us/red_hat_build_of_keycloak/24.0/html-single/server_administration_guide/index#password_blacklist>

PR #225, opened 2024-05-14
#224:  keycloak_quarkus: Add support for policy files

hwo-wd:


guidograzioli (changes requested):
Changeset good, but missing the argument spec (https://github.com/ansible-middleware/keycloak/blob/main/roles/keycloak_quarkus/meta/argument_specs.yml) definition for the two new parameters. For myself, need to verify why the linter rule didn't report same

guidograzioli (approved):

PR #227, opened 2024-05-14
`proxy-header` enhancement

### LINKED ISSUE #226: KC>22: Don't set (deprecated) `proxy` setting when `proxy-headers` are set
KC startup looks as follows:

> WARN  [org.keycloak.quarkus.runtime.cli.Picocli] (main) The following used options or option values are DEPRECATED and will be removed in a future release:
        - proxy: Use proxy-headers.

and the keycloak.conf:

```text
# Deprecated Proxy configuration
proxy=edge
# Proxy
proxy-headers=xforwarded
```

-> `proxy` should only pe set when `keycloak_quarkus_proxy_headers` are not set to get rid of the warning

hwo-wd:
Deprecated `proxy` setting will only be emitted in keycloak.conf if `keycloak_quarkus_proxy_headers` is not set.

Fix #226

### DIFF
--- a/roles/keycloak_quarkus/meta/argument_specs.yml
+++ b/roles/keycloak_quarkus/meta/argument_specs.yml
@@ -456,7 +456,7 @@ argument_specs:
     downstream:
         options:
             rhbk_version:
-                default: "24.0.4"
+                default: "24.0.3"
                 description: "Red Hat Build of Keycloak version"
                 type: "str"
             rhbk_archive:
--- a/roles/keycloak_quarkus/meta/argument_specs.yml
+++ b/roles/keycloak_quarkus/meta/argument_specs.yml
@@ -483,6 +483,10 @@ argument_specs:
                 default: false
                 description: "Perform an offline install"
                 type: "bool"
+            keycloak_quarkus_show_deprecation_warnings:
+                default: true
+                description: "Whether deprecation warnings should be shown"
+                type: "bool"
             rhbk_service_name:
                 default: "rhbk"
                 description: "systemd service name for Red Hat Build of Keycloak"
--- a/roles/keycloak_quarkus/tasks/deprecations.yml
+++ b/roles/keycloak_quarkus/tasks/deprecations.yml
@@ -34,3 +34,20 @@
 
     - name: Flush handlers
       ansible.builtin.meta: flush_handlers
+
+# https://access.redhat.com/documentation/en-us/red_hat_build_of_keycloak/24.0/html-single/upgrading_guide/index#deprecated_literal_proxy_literal_option
+- name: Check deprecation of keycloak_quarkus_proxy_mode
+  delegate_to: localhost
+  run_once: true
+  when:
+    - keycloak_quarkus_proxy_mode is defined
+    - keycloak_quarkus_proxy_headers is defined and keycloak_quarkus_proxy_headers | length == 0
+    - keycloak_quarkus_version.split('.') | first | int >= 24
+  changed_when: true
+  ansible.builtin.set_fact:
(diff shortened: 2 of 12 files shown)


roles/keycloak_quarkus/templates/keycloak.conf.j2
```diff
@@ -69,13 +69,12 @@ cache-config-file=cache-ispn.xml
 {% endif %}
 {% endif %}
 
-{% if keycloak_quarkus_proxy_mode is defined and keycloak_quarkus_proxy_mode != "none" %}
-# Deprecated Proxy configuration
-proxy={{ keycloak_quarkus_proxy_mode }}
-{% endif %}
 {% if keycloak_quarkus_proxy_headers is defined and keycloak_quarkus_proxy_headers != "none" %}
```
guidograzioli, line 72: Problem here coming from somewhere else; it seems keycloak_quarkus_proxy_headers is either "", "forwarded" or "xforwarded", but never "none". Can you change here to keycloak_quarkus_proxy_mode != ""

guidograzioli (changes requested):

roles/keycloak_quarkus/templates/keycloak.conf.j2
hwo-wd: good catch, thx; added proper validation

roles/keycloak_quarkus/meta/argument_specs.yml
```diff
@@ -412,7 +412,7 @@ argument_specs:
     downstream:
         options:
             rhbk_version:
-                default: "24.0.4"
+                default: "24.0.3"
```
guidograzioli, line 415: thx!

guidograzioli (approved):
LGTM

guidograzioli:
There is another earlier problem: keycloak_quarkus_proxy_headers (https://github.com/ansible-middleware/keycloak/blob/main/roles/keycloak_quarkus/meta/argument_specs.yml#L329) is defined as default '', but the default is missing from defaults/main.yml, making it essentially a variable.

Can you take care of that here? (adding the "" default (preferred), or otherwise dropping it from arg_specs, and handling the undefined case)

guidograzioli:
I do not really understand why the idempotency issue happens here (and not in previous builds); I think we have to drop the changed_when: true from [here](url) (which i added about a week ago to have orange logs instead of red...)

roles/keycloak_quarkus/tasks/deprecations.yml
```diff
@@ -34,3 +34,20 @@
...
+
+# https://access.redhat.com/documentation/en-us/red_hat_build_of_keycloak/24.0/html-single/upgrading_guide/index#deprecated_literal_proxy_literal_option
+- name: Check deprecation of keycloak_quarkus_proxy_mode
+  delegate_to: localhost
+  run_once: true
+  when:
+    - keycloak_quarkus_proxy_mode is defined
+    - keycloak_quarkus_proxy_headers is defined and keycloak_quarkus_proxy_headers | length == 0
+    - keycloak_quarkus_version.split('.') | first | int >= 24
+  changed_when: true
```
guidograzioli, line 46: I propose, for the sake of idempotency in tests, to change this to:
`changed_when: keycloak_quarkus_deprecation_warnings`
keycloak_quarkus_deprecation_warnings is default true, but we set it to false in all molecule tests

wdyt?

guidograzioli (approved):

hwo-wd:
@guidograzioli this is still WIP, still trying to figure out how to make it work, please don't merge just yet

Issue #221, opened 2024-05-12
Hardcoded values in roles/keycloak_quarkus/vars/main.yml


dmitriytsokolov:
##### SUMMARY

Hi! I'm setting up a keycloak using this middleware and have a problem. I'm using https with self-signed certificate. Everything works fine except "healthcheck" step, because "http" is hardcoded in the file. I know I can set hash to merge in ansible config, but it would be great to have this part parametrized as well as this part: /realms/master/.well-known/openid-configuration

Thanks in advance

##### ISSUE TYPE
 - Feature Idea

PR #230, opened 2024-05-14
Close #182, #221: improve restart handler logic

### LINKED ISSUE #182: keycloak_quarkus role restart all node at the same times
##### SUMMARY
If a configuration change, actually the keycloak_quarkus role restart all node at the same time.
So the keycloak service is down for this (small) period.

It's also introduce a problem if try to apply a bad configuration : 
all node will be change and restart so the whole keycloak service will be down until a correction is done
as exemple, if i change the ssl file to a not existing file (or permission issue or ... ) with : 
```
keycloak_quarkus_key_file: "/etc/ssl/private/not_existing_file.key.pem"
```
All keycloak node will be down, so the service will be down ...

I think, we should, at least, introduce a `throttle` or `forks` on restart service 
Or even better apply change on first node before other node

Personally i like how restart have been implemented on this role : https://github.com/mrlesmithjr/ansible-mariadb-galera-cluster/blob/master/tasks/setup_cluster.yml : 
- it's not use notify but `register: "_mariadb_galera_cluster_reconfigured"`
- then apply change to first node only : 
```
- name: setup_cluster | cluster rolling restart - apply config changes (first node)
  ansible.builtin.include_tasks: manage_node_state.yml
```
- then (if successfully restart) restart other : 
```
- name: setup_cluster | cluster rolling restart - apply config changes (other nodes)
  ansible.builtin.include_tasks: manage_node_state.yml
```
So with this solution it's resolving both issue i've describe earlier

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
core 2.14.1
```

##### COLLECTION VERSION

```
middleware_automation.keycloak 2.1.0  

```

##### STEPS TO REPRODUCE

- change a configuration like on `keycloak_quarkus_frontend_url`
- run the playbook with the role `middleware_automation.keycloak.keycloak_quarkus`

##### EXPECTED RESULTS
- all keycloak service need to be restarted one after the other 

##### ACTUAL RESULTS
- all keycloak service will restart at the same time and without any error handling

##### ADDITIONAL INFORMATION

also, i think `start.yml` and `restart.yml` need to be merge ( `state: restarted` is always doing a `state: started` and more ) and the "Wait until {{ keycloak.service_name }} becomes active {{ keycloak.health_url }}" need also to be used on the restart behavior.

hwo-wd:
This does restart sequentially (at least a single node gets restarted, then the others);

@guidograzioli:
* [ ] not sure whether you want to check the health differently (`wait_for_healthy` as you stated in #182), but I guess this is the right PR to tackle all these.
* [x] Re: #221:
  > [..] but it would be great to have this part parametrized as well as this part: /realms/master/.well-known/openid-configuration

  This does not seem fruitful since the `master` realm needs to exist always afaik, so this is the perfect place to check against, wdyt?
  
* [x] Re: CI: Not sure why the molecule tests fail, any pointers? Surely it's related to the `health_url`, but I don't see how...

### DIFF
--- a/roles/keycloak_quarkus/tasks/restart.yml
+++ b/roles/keycloak_quarkus/tasks/restart.yml
@@ -1,9 +1,38 @@
 ---
-- name: "Restart and enable {{ keycloak.service_name }} service"
+- name: Ensure only one service at a time gets rebooted, to ensure replication of distributed ispn caches
   throttle: 1
-  ansible.builtin.systemd:
-    name: keycloak
-    enabled: true
-    state: restarted
-    daemon_reload: true
-  become: true
+  block:
+    - name: "Restart and enable {{ keycloak.service_name }} service on first host"
+      ansible.builtin.systemd:
+        name: "{{ keycloak.service_name }}"
+        enabled: true
+        state: restarted
+        daemon_reload: true
+      become: true
+      delegate_to: "{{ ansible_play_hosts | first }}"
+      run_once: true
+
+    - name: "Wait until {{ keycloak.service_name }} service becomes active {{ keycloak.health_url }}"
+      ansible.builtin.uri:
+        url: "{{ keycloak.health_url }}"
+      register: keycloak_status
+      until: keycloak_status.status == 200
+      retries: 25
+      delay: 10
+      delegate_to: "{{ ansible_play_hosts | first }}"
+      run_once: true
+
+    - name: Pause to give distributed ispn caches time to (re-)replicate back onto first host
+      ansible.builtin.pause:
+        seconds: 15
+      when:
+        - rhbk_ha_enabled
+
+    - name: "Restart and enable {{ keycloak.service_name }} service on all other hosts"
+      ansible.builtin.systemd:
+        name: "{{ keycloak.service_name }}"
+        enabled: true
+        state: restarted
+        daemon_reload: true
+      become: true
+      when: inventory_hostname != ansible_play_hosts | first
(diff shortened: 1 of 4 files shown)


guidograzioli:
Change is going is right direction, but I have some additional idea on how to handle the restart (ie. include at least a couple of strategies, and allow for plugging in own restart behaviour). Happy to co-author this PR?

roles/keycloak_quarkus/tasks/restart.yml
```diff
@@ -1,9 +1,38 @@
...
+      retries: 25
+      delay: 10
+      delegate_to: "{{ ansible_play_hosts | first }}"
+      run_once: true
+
+    - name: Pause to give distributed ispn caches time to (re-)replicate back onto first host
+      ansible.builtin.pause:
+        seconds: 15
+      when:
+        - rhbk_ha_enabled
```
guidograzioli, line 29: `keycloak_quarkus_`

guidograzioli:
> This does not seem fruitful since the master realm needs to exist always afaik, so this is the perfect place to check against, wdyt?

I totally agree, however parameterizing that urlpath does not harm us either (keeping the current as default of course)

hwo-wd:
> Change is going is right direction, but I have some additional idea on how to handle the restart (ie. include at least a couple of strategies, and allow for plugging in own restart behaviour). Happy to co-author this PR?



> > This does not seem fruitful since the master realm needs to exist always afaik, so this is the perfect place to check against, wdyt?
> 
> I totally agree, however parameterizing that urlpath does not harm us either (keeping the current as default of course)

Ok, I've added an `keycloak_quarkus_health_check_url_path` option

PR #231, opened 2024-05-15
Restart handler strategy behaviour

### LINKED ISSUE #182: keycloak_quarkus role restart all node at the same times

### LINKED ISSUE #221: Hardcoded values in roles/keycloak_quarkus/vars/main.yml

guidograzioli:
Builds on and supersedes #230; a directory for restart strategies is provided with three implementations: `none` (nothing is restarted), `serial` (restart hosts in a serial fashion), `serial_then_parallel` (restart first node, verify health url, proceed with the rest in parallel). Restart health check can be mix-and-matched with the wait_for parameters.

The default strategy file, imported from the restart handler, is 'serial_then_parallel'; user can use that, choose one of the other too, or specify their custom taskfile (in a relative path of the calling playbook).

* `serial` is the only viable restart behaviour on keycloak with default embedded infinispan caches (distributed-caches with owners = 2 implies a single point of failure).
* `serial_then_parallel` is for the case of remote infinispan caches, where liveliness of the first node to be restarted allows the other keycloak node to be restarted in parallel

The molecule scenario `quarkus_ha` tests a cluster of two keycloak instances with a shared postgresql instance, restart in serial strategy

New parameters:

|`keycloak_quarkus_restart_strategy`| Strategy task file for restarting in HA (one of provided restart/['serial.yml','none.yml','serial_then_parallel.yml']) or path to file when providing custom strategy | `restart/serial.yml` |
|`keycloak_quarkus_restart_health_check`| Whether to wait for successful health check after restart | `{{ keycloak_quarkus_ha_enabled }}` |
|`keycloak_quarkus_restart_health_check_delay`| Seconds to let pass before starting healch checks | `10` |
|`keycloak_quarkus_restart_health_check_reries`| Number of attempts for successful health check before failing | `25` |
|`keycloak_quarkus_restart_pause`| Seconds to wait between restarts in HA strategy | `15` |

Fix #182
Fix #221


guidograzioli:
@hwo-wd please review; if we agree on this coauthored PR, I'll find another pair of eye to do the formal review

roles/keycloak_quarkus/tasks/restart/serial.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: "Restart services in serial, with optional healtch check (keycloak_quarkus_restart_health_check)"
+  throttle: 1
+  loop: "{{ ansible_play_hosts }}"
+  block:
+    - name: "Restart and enable {{ keycloak.service_name }} service on first host"
```
hwo-wd, line 6: ... service on {{ item }}"

roles/keycloak_quarkus/tasks/restart.yml
```diff
@@ -1,38 +1,17 @@
...
-      ansible.builtin.systemd:
-        name: "{{ keycloak.service_name }}"
-        enabled: true
-        state: restarted
-        daemon_reload: true
-      become: true
-      when: inventory_hostname != ansible_play_hosts | first
+- name: "Wait until {{ keycloak.service_name }} service becomes active {{ keycloak.health_url }}"
+  ansible.builtin.uri:
+    url: "{{ keycloak.health_url }}"
```
hwo-wd, line 12: with `keycloak.health_url` being defined the way it is and thinking about a production scenario (ha-enabled, fronted by reverse proxy etc.) this should give a 200 ok almost immediatelly - independent of the return status of the current host.
Not saying this is bad, but possible alternatives could be:
* template `health_url` to include the rebooted host and checking against the actual node; downside: the ansible host not necessarily needs to reach the host on the given port, depending on the FW configuration
* wait and check logs whether the service has started

roles/keycloak_quarkus/meta/argument_specs.yml
```diff
@@ -425,6 +431,20 @@ argument_specs:
                 default: true
                 description: "Allow the option to ignore invalid certificates when downloading JDBC drivers from a custom URL"
                 type: "bool"
+            keycloak_quarkus_restart_health_check:
```
hwo-wd, line 434: please add these as docs to `readme.md`

roles/keycloak_quarkus/meta/argument_specs.yml
```diff
@@ -425,6 +431,20 @@ argument_specs:
...
+                default: "{{ keycloak_quarkus_ha_enabled }}"
+                description: "Whether to wait on successful health check after restart"
+                type: "bool"
+            keycloak_quarkus_restart_strategy:
+                description: >
+                  Strategy task file for restarting in HA, one of [ 'serial', 'none', 'verify_first' ] below, or path to
```
hwo-wd, line 440: Strategy task file for restarting if `keycloak_quarkus_restart_health_check==True`,...

roles/keycloak_quarkus/tasks/restart/serial.yml
```diff
@@ -0,0 +1,8 @@
...
+      ansible.builtin.include_tasks: ../restart.yml
+      delegate_to: "{{ item }}"
```
hwo-wd, line 8: I guess we might need the pause task here as well, otherwise the health check returns 200 OK quite early (since the 2nd node is up) -> we restart the 2nd service -> ispn cache lost; considering this is the default strategy (if ha enabled), we should make this the most failsafe, ableit the slowest

```
    - name: Pause to give distributed ispn caches time to (re-)replicate back onto first host
      ansible.builtin.pause:
        seconds: "{{ keycloak_quarkus_restart_pause }}"
      when:
        - keycloak_quarkus_ha_enabled
 ```

File: roles/keycloak_quarkus/tasks/restart/verify_first.yml
Comment (hwo-wd) [inline roles/keycloak_quarkus/tasks/restart/verify_first.yml:1]: not sure about `verify_first` naming, the longer I think about it the more it confuses me;
how about `restart_first_verify_restart_rest` - not a nice name either, but more descriptive

**Review by hwo-wd — CHANGES_REQUESTED** [review]
Mostly nitpicking

File: roles/keycloak_quarkus/tasks/restart.yml (hunk 2 above)
Comment (guidograzioli) [reply roles/keycloak_quarkus/tasks/restart.yml:12]: My intent here was indeed to contact localhost (localhost from delegate_to being the restarted node); do you mean in production scenario keycloak_quarkus_host would change from localhost? Ideally it should be keycloak_quarkus_frontend_url that takes the load balanced / reverse proxied domainname (no?).

File: roles/keycloak_quarkus/tasks/restart/serial.yml (hunk 5 above)
Comment (guidograzioli) [reply roles/keycloak_quarkus/tasks/restart/serial.yml:8]: That's very true, I'll move the sleeping out of the strategy and directly inside the main restart.yml block

File: roles/keycloak_quarkus/tasks/restart/verify_first.yml
Comment (guidograzioli) [reply roles/keycloak_quarkus/tasks/restart/verify_first.yml:1]: restart is implied by the restart/ subdirectory... how about first_then_parallel ?

Comment (guidograzioli) [reply roles/keycloak_quarkus/tasks/restart/verify_first.yml:1]: let me expand on this, I probably jumped to fast to code with out design notes.

-  for HA clusters with embedded infinispan (on defaults, so with distributed-caches with 2 owners), we can tolerate just one node to go down at a time, so serial is the really only one strategy that works
- restarting a single node, verifying its liveliness, then parallelize all other node restarts is for remote caches. When the first node goes down, the others preserve the service; when the first is up again, it will take the traffic and all others can restart. Ideally that "first" should become a parameter itself, could be "2", "owners-1" or even, one day, "one per region".

File: roles/keycloak_quarkus/tasks/restart.yml (hunk 2 above)
Comment (hwo-wd) [reply roles/keycloak_quarkus/tasks/restart.yml:12]: you're right, I implied that `keycloak_quarkus_host` (=localhost, by default) ==`rhbk_frontend_url` (=load balanced) which is not the case -- all good, sry for the confusion

File: roles/keycloak_quarkus/tasks/restart/serial_then_parallel.yml [hunk 6]
```diff
@@ -0,0 +1,37 @@
+---
+- name: Verify first restarted service with health URL, then rest restart in parallel
+  block:
+    - name: "Restart and enable {{ keycloak.service_name }} service on first host"
```
hwo-wd, line 4: we should re-use `restart.yml` here with `delegate_to: "{{ ansible_play_hosts | first }}"` and `run_once: true`, i.e. replacing the next three tasks with this include instead

roles/keycloak_quarkus/handlers/main.yml
```diff
@@ -7,7 +7,8 @@
   ansible.builtin.include_tasks: bootstrapped.yml
   listen: bootstrapped
 - name: "Restart {{ keycloak.service_name }}"
-  ansible.builtin.include_tasks: restart.yml
+  ansible.builtin.include_tasks:
+    file: "{{ keycloak_quarkus_restart_strategy if keycloak_quarkus_ha_enabled else 'restart.yml' }}"
   listen: "restart keycloak"
```
hwo-wd, line 12: the linter seems to prefer `Restart keycloak` here

guidograzioli: That's new.. and kind of non-sense to me https://github.com/ansible/ansible-lint/issues/4168

roles/keycloak_quarkus/tasks/restart/serial_then_parallel.yml
guidograzioli: not really; restart.yml can have the health check turned off, while in there is enforced

hwo-wd: your right, but still: e.g. introducing `keycloak_quarkus_internal_enforce_health_check` and setting it to true here would lower the maintenance burden on the long run -- your call, of course

guidograzioli: mmh after second thought, yeah you're right, let me amend that

sabre1041 (approved):
LGTM

guidograzioli:
@hwo-wd I believe all your notes were addressed, and we got the green flag from Andy; if you have other commits to push or comments please do; or, I'll proceed and merge. This has been pretty epic :)

hwo-wd:
lgtm, thanks for the journey :)

PR #252, opened 2024-12-05
keycloak_quarkus: Add theme cache invalidation handler

hwo-wd:


roles/keycloak_quarkus/tasks/install.yml
```diff
@@ -218,7 +218,7 @@
   become: true
   loop: "{{ keycloak_quarkus_providers }}"
   when: item.url is defined and item.url | length > 0
-  notify: "{{ ['rebuild keycloak config', 'restart keycloak'] if not item.restart is defined or item.restart else [] }}"
+  notify: "{{ ['invalidate keycloak theme cache', 'rebuild keycloak config', 'restart keycloak'] if not item.restart is defined or item.restart else ['invalidate keycloak theme cache'] }}"
```
hwo-wd, line 221: I'm undecided whether the `else` part makes sense: on the one hand I don't like the idea to delete a cache a running (production) instance might be running w/o seeing a restart; still, I added it for discussion purposes, but overall I tend to go the `else []` route.
Any thoughts, @guidograzioli? -- thanks

guidograzioli: I am not accustomed to keycloak themes, are we talking about a server-side freemarker cache here? Or deleting that file server-side will serve the invalidate cache header to the clients? IN the former case I'd `else[]`, otherwise I'd notify the cache invalidation. But own .02c here

guidograzioli: Changeset is however approved :+1 with whatever version of the else you drop here.

guidograzioli (approved):
LGTM

roles/keycloak_quarkus/tasks/install.yml
hwo-wd: Thanks for your input; it's the former: it's a (preprocessed?) and gzipped version of the template delivered by the server, so I went the `else []` route

Issue #256, opened 2024-12-20
Some steps don't skip certificate check


NilsDeckert:
##### SUMMARY

Creating a new realm on a keycloak instance with self-signed certificates fails with `SSL: CERTIFICATE_VERIFY_FAILED`.
Certificate checks are skipped in most steps, hence I assume that skipping and not skipping within the same task is not done on purpose.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible [core 2.18.0]
  config file = /home/nils/code/my_project/ansible-playbooks/ansible.cfg
  configured module search path = ['/home/nils/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.12/site-packages/ansible
  ansible collection location = /home/nils/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.12.7 (main, Oct  1 2024, 11:15:50) [GCC 14.2.1 20240910] (/usr/bin/python)
  jinja version = 3.1.4
  libyaml = True
```

##### COLLECTION VERSION

```
 /home/nils/.ansible/collections/ansible_collections
Collection                               Version
---------------------------------------- -------
fedora.linux_system_roles                1.93.0 
middleware_automation.common             1.2.2  
middleware_automation.keycloak           2.4.3  
nextcloud.admin                          2.0.0  
```

##### STEPS TO REPRODUCE

```yaml
- name: Create a new realm
  hosts: keycloak
  remote_user: ansible
  become: true
  gather_facts: true

  roles:
    - role: middleware_automation.keycloak.keycloak_realm
      vars:
        keycloak_no_log: false
        keycloak_admin_user: "admin"
        keycloak_admin_password: "admin"
        keycloak_url: "https://{{ ansible_default_ipv4.address }}:8443"
        keycloak_context: ""
        keycloak_realm: "new-realm"
```

##### EXPECTED RESULTS

Creation of a new realm 'new-realm'.

##### ACTUAL RESULTS

```
TASK [middleware_automation.keycloak.keycloak_realm : Generate keycloak auth token] ***************************************************************************************************************************************************
ok: [123.123.123.123]

TASK [middleware_automation.keycloak.keycloak_realm : Determine if realm exists] ******************************************************************************************************************************************************
[...]

PR #257, opened 2024-12-20
Skip certificate checking

NilsDeckert:
See #256.
Most, but not all steps in the `keycloak_realm` tasks skip checking SSL certificates.

Since e.g. in `roles/keycloak_realm/tasks/main.yml` the first step `Generate keycloak auth token` has set
 `validate_certs: false`, but the next step `Determine if realm exists` does not, I assume this was a mistake?
 
 I briefly checked the other tasks in `roles/keycloak_realm/tasks/` and added `validate_certs: false` where it was missing.
 With those changes, the tasks now work for my setup with self-signed certificates.

### DIFF
--- a/roles/keycloak_realm/tasks/main.yml
+++ b/roles/keycloak_realm/tasks/main.yml
@@ -15,6 +15,7 @@
   ansible.builtin.uri:
     url: "{{ keycloak_url }}{{ keycloak_context }}/admin/realms/{{ keycloak_realm }}"
     method: GET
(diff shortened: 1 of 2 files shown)


guidograzioli (approved):
lgtm

guidograzioli:
I am merging this PR (thanks for contributing), with a note: the whole _realm role with likely undergo breaking changes / a total rewrite in the coming weeks. YMMV

PR #260, opened 2025-02-16
AMW-365 Keycloak collection GitHub action molecule pipelines are breaking because of sudo permission issue

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/AMW-365


guidograzioli:
Lets keep the branch in case other problems arise, but downgrading the github runner to 22.04 (instead of 24.04 / latest) resolves the issue

PR #269, opened 2025-04-13
Update example playbooks to use new `bootstrap` var

Preskton:
When running the quarkus role using the example provided in the repo, it fails with:

```
TASK [middleware_automation.keycloak.keycloak_quarkus : Validating arguments against arg spec 'main'] *********************************************************************************************************************
fatal: [91.99.14.247]: FAILED! => {"argument_errors": ["argument 'keycloak_quarkus_bootstrap_admin_password' is of type <class 'NoneType'> and we were unable to convert to str:
```

This replaces the deprecated `keycloak_admin_password` with `keycloak_quarkus_bootstrap_admin_password`.

### DIFF
--- a/playbooks/keycloak_quarkus.yml
+++ b/playbooks/keycloak_quarkus.yml
@@ -2,7 +2,7 @@
 - name: Playbook for Keycloak X Hosts with HTTPS enabled
   hosts: all
   vars:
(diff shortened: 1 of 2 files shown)


guidograzioli (approved):
Thanks for your contribution!

PR #271, opened 2025-04-16
Rename and honour parameter `keycloak_quarkus_http_host`

### LINKED ISSUE #242: Missing possibility to configure http-host
I cannot find the possibility to configure the parameter http-host in the keycloak.conf .
Have I overlooked it or is it missing?
Kind regards

guidograzioli:
Parameter rename and add in configuration since it was missing:

| Name                                           | Replacement (requires manual adaptation)   | Reason                                                                                               |
| ---------------------------------------------- | ----------------------------------------------- | ----------------------------------------------------------- |
| `keycloak_quarkus_bind_address`         | `keycloak_quarkus_http_host`  | [HTTP configuration](https://www.keycloak.org/server/all-config#category-http)  |

`keycloak_quarkus_bind_address` remains as deprecated, when used a deprecation warning will be emitted and the variable will be used.

Labelled as bugfix and not breaking change even if it contains a rename, because the value would not be used when set.

Fix #242


hwo-wd (approved):
lgtm

PR #274, opened 2025-04-18
Load environment vars during kc rebuild

### LINKED ISSUE #273: Is there any way of disabling KC features?
### Discussed in https://github.com/ansible-middleware/keycloak/discussions/272

<div type='discussions-op-text'>

<sup>Originally posted by **danielmatthews84** April 17, 2025</sup>
I can't see any way of disabling KC features using the `keycloak_quarkus` role. There doesn't appear to be any vars for this and I have tried setting `keycloak_quarkus_additional_env_vars` as below but these changes do not trigger a rebuild and the KC service fails to start
```
keycloak_quarkus_additional_env_vars:
    - key: KC_FEATURES_DISABLED
      value: impersonation,kerberos
```

Am I missing something or is this not supported by this role?</div>

guidograzioli:
A change to environment file now triggers a server rebuild.

Fix #273

### DIFF
--- a/roles/keycloak_quarkus/tasks/rebuild_config.yml
+++ b/roles/keycloak_quarkus/tasks/rebuild_config.yml
@@ -2,9 +2,6 @@
 # cf. https://www.keycloak.org/server/configuration#_optimize_the_keycloak_startup
 - name: "Rebuild {{ keycloak.service_name }} config"
   ansible.builtin.shell: | # noqa blocked_modules shell is necessary here
-    {{ keycloak.home }}/bin/kc.sh build
-  environment:
-    PATH: "{{ keycloak_quarkus_java_home | default(keycloak_quarkus_pkg_java_home, true) }}/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
-    JAVA_HOME: "{{ keycloak_quarkus_java_home | default(keycloak_quarkus_pkg_java_home, true) }}"
+    env -i bash -c "set -a ; source {{ keycloak_quarkus_sysconf_file }} ; {{ keycloak.home }}/bin/kc.sh build "
   become: true
   changed_when: true
--- a/roles/keycloak_quarkus/templates/keycloak-sysconfig.j2
+++ b/roles/keycloak_quarkus/templates/keycloak-sysconfig.j2
@@ -5,9 +5,9 @@ KC_BOOTSTRAP_ADMIN_PASSWORD='{{ keycloak_quarkus_bootstrap_admin_password }}'
 {% else %}
 {{ keycloak.bootstrap_mnemonic }}
 {% endif %}
-PATH={{ keycloak_quarkus_java_home | default(keycloak_sys_pkg_java_home, true) }}/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
-JAVA_HOME={{ keycloak_quarkus_java_home | default(keycloak_sys_pkg_java_home, true) }}
-JAVA_OPTS='{{ keycloak_quarkus_java_opts }}'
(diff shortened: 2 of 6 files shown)


roles/keycloak_quarkus/tasks/rebuild_config.yml
```diff
@@ -2,9 +2,6 @@
 # cf. https://www.keycloak.org/server/configuration#_optimize_the_keycloak_startup
 - name: "Rebuild {{ keycloak.service_name }} config"
   ansible.builtin.shell: | # noqa blocked_modules shell is necessary here
-    {{ keycloak.home }}/bin/kc.sh build
-  environment:
-    PATH: "{{ keycloak_quarkus_java_home | default(keycloak_quarkus_pkg_java_home, true) }}/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
-    JAVA_HOME: "{{ keycloak_quarkus_java_home | default(keycloak_quarkus_pkg_java_home, true) }}"
+    env -i bash -c "set -a ; source {{ keycloak_quarkus_sysconf_file }} ; {{ keycloak.home }}/bin/kc.sh build "
```
hwo-wd, line 5: is it:
* safe to assume, that `bash` is installed? `sh -c` would be sufficing too, wouldn't it? I'm thinking of alpine images and this kind, even though it's unlikely that this role will be used
* the `set -a` is intentional? -> I'd go for a `set -e` instead maybe?
  > -a  Mark variables which are modified or created for export.

roles/keycloak_quarkus/templates/keycloak-sysconfig.j2
```diff
@@ -7,7 +7,7 @@ KC_BOOTSTRAP_ADMIN_PASSWORD='{{ keycloak_quarkus_bootstrap_admin_password }}'
 {% endif %}
 PATH={{ keycloak_quarkus_java_home | default(keycloak_sys_pkg_java_home, true) }}/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 JAVA_HOME={{ keycloak_quarkus_java_home | default(keycloak_sys_pkg_java_home, true) }}
```
hwo-wd, line 9: since you've escaped `JAVA_OPTS`, I'm just thinking whether all should be escaped, helps with spaces and issues like that, wdyt?

hwo-wd (commented):
some small comments from my side

roles/keycloak_quarkus/tasks/rebuild_config.yml
guidograzioli: Actually the `set -a` is necessary, as it marks the envvars loaded from the `source` to be passed to the next command executed in the child shell process. As for bash, it is quite a requirement for all middleware_automation collections.... but it's a good point I'll add to the installed_packages

roles/keycloak_quarkus/templates/keycloak-sysconfig.j2
guidograzioli: not sure, I quoted because JVM args could (actually likely will) start with an hyphen, and that would not work when loaded with `source`. Are paths safe? wdyt?

hwo-wd: tbh, I'd add them: it doesn't hurt and if it doesn't break a molecule, we should be good
Refs:
* <https://unix.stackexchange.com/a/26059> (esp. the last paragraph)
* <https://stackoverflow.com/a/12902280>

so `JAVA_HOME` and `PATH`, but using `"` in case one wants to use env vars for whatever reasons :)

guidograzioli: ok, I'm convinced! :)

hwo-wd (approved):

guidograzioli:
@hwo-wd release 3.0? this is the changelog:
https://collections.ansiblemiddleware.com/keycloak/main/CHANGELOG.html#v3-0-0-devel

PR #281, opened 2025-05-12
New parameter to set the jgroups host IP address

jonathanspw:
This is useful if the default route does not
represent the network you want/need to use for
cluster communication.

### DIFF
--- a/molecule/debian/prepare.yml
+++ b/molecule/debian/prepare.yml
@@ -10,3 +10,8 @@
           # - openjdk-21-jdk-headless # this is not available in ghcr.io/hspaans/molecule-containers:debian-11 (neither in debian-12) since the images are using outdated package sources
           - openjdk-17-jdk-headless
         state: present
+    - name: "Install iproute2"
+      ansible.builtin.apt:
+        name:
+          - iproute2
+        state: present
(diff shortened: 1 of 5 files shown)


guidograzioli (changes requested):
Thank you for your contribution; the changeset is ok, but, can you please:
- add the parameter definition in meta/argument_specs (surprised the linter didn't complain? 
- update molecule/quarkus_ha to make use of the parameter?

jonathanspw:
I've updated `meta/argument_specs.yml` to add the definition.

I'm not sure what to update for molecule...I see most of it is symlinks to the actual role - and all of the files appear to be using the new var.

I did notice that the debian tests on molecule need the `iproute2` package in order for `ansible_default_ipv4.address` to be available when it is processed from `defaults/main.yml` so I added an install task for it in `prepare.yml`.

guidograzioli:
Alright; fyi, I meant to set hostvars [here](https://github.com/ansible-middleware/keycloak/blob/main/molecule/quarkus_ha/molecule.yml#L63) for the hosts forming the cluster, but no worries I'll add in another change since I am renewing the Ci a little bit anyway.

guidograzioli (approved):
LGTM
Thanks for your contribution!

PR #285, opened 2025-05-20
Run config rebuild after SPI providers update

hwo-wd:
26.2 is stricter when it comes to validating its optimized config and thus refuses to start in case a SPI was deployed w/o properly having invoked `kc.sh build` first (related: https://github.com/keycloak/keycloak/issues/37770)


guidograzioli (approved):
lgtm

guidograzioli:
Merge at will, I reworded the title of the PR a little for the changelog

PR #287, opened 2025-05-29
Session storage / distributed caches

guidograzioli:
https://www.keycloak.org/server/caching
https://www.keycloak.org/2024/12/storing-sessions-in-kc26


guidograzioli:
needs https://github.com/ansible-middleware/infinispan/pull/49

PR #302, opened 2025-07-01
keycloak_realm: federation default provider type should be a string

tinsjourney:


guidograzioli (approved):
LGTM, thanks for your contribution!

Issue #261, opened 2025-03-07
manage_client_roles uses client.name as client_id


aphick:
https://github.com/ansible-middleware/keycloak/blob/910a2aa5d43a5003cab16deb68e65e7bd5fe20dc/roles/keycloak_realm/tasks/manage_client_roles.yml#L5

Shouldn't this be `{{ client.client_id }}`?

RanabirChakraborty:
Check here how the variables are placed: https://github.com/ansible-middleware/keycloak/blob/910a2aa5d43a5003cab16deb68e65e7bd5fe20dc/molecule/default/converge.yml#L37-L43, and you need to pass the variables like this only.

Please let me know if it resolves the issue.

PR #307, opened 2025-07-16
manage_client_roles.yml: use "client.id" instead of "client.name" to fix client role creation.

sgoericke:
I faced an issue today that client role could not be added. Turned out that by accident the "client.name" variable was used instead of "client.id" in the task.


RanabirChakraborty (approved):
LGTM

RanabirChakraborty:
GitHub issue: https://github.com/ansible-middleware/keycloak/issues/261

PR #308, opened 2025-07-21
Fix config_key_store_file description to match variable name

tinsjourney:
Small fix to update description for keycloak_quarkus_config_key_store_file variable in order to match the password variable name.


hwo-wd (commented):
lgtm

PR #325, opened 2026-04-23
Fix keycloak_context default from /auth to empty string for Quarkus-based Keycloak

paulomenon:
## Summary

- Change the default value of `keycloak_context` from `/auth` to `''` (empty string) in the `keycloak_realm` role
- Update `argument_specs.yml` and `README.md` documentation to reflect the new default
- Remove the now-unnecessary `keycloak_context: ''` overrides from all 4 molecule converge files (`quarkus`, `debian`, `quarkus_devmode`, `default`)

## Problem

The `/auth` context path was used by legacy **WildFly-based Keycloak** (pre-v17). Since the move to Quarkus (Keycloak 17+), the context path is empty by default. The same applies to **Red Hat Build of Keycloak (RHBK)**.

With the current default of `/auth`, all REST API URLs are built incorrectly (e.g. `http://localhost:8080/auth/realms/master/...` instead of `http://localhost:8080/realms/master/...`), forcing every user of Quarkus-based Keycloak/RHBK to explicitly pass:

```yaml
keycloak_context: ""
```

This is a common source of confusion — the molecule tests themselves already had to work around it by setting `keycloak_context: ''` in every converge file.

## Fix

- `roles/keycloak_realm/defaults/main.yml`: `keycloak_context: ''`
- `roles/keycloak_realm/meta/argument_specs.yml`: `default: ""`
- `roles/keycloak_realm/README.md`: updated docs table
- `molecule/*/converge.yml`: removed redundant `keycloak_context: ''` overrides (4 files)

Users still on legacy WildFly-based Keycloak can set `keycloak_context: /auth` explicitly.

## Test plan

- [ ] Existing molecule tests should pass without the explicit `keycloak_context: ''` override
- [ ] Verify Quarkus-based Keycloak realm creation works out of the box without setting context
- [ ] Verify legacy WildFly users can still override with `keycloak_context: /auth`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/molecule/debian/converge.yml
+++ b/molecule/debian/converge.yml
@@ -13,7 +13,6 @@
     - role: keycloak_quarkus
     - role: keycloak_realm
       keycloak_url: "{{ keycloak_quarkus_hostname }}"
-      keycloak_context: ''
       keycloak_admin_user: "{{ keycloak_quarkus_bootstrap_admin_user }}"
       keycloak_admin_password: "{{ keycloak_quarkus_bootstrap_admin_password }}"
       keycloak_client_users:
--- a/molecule/default/converge.yml
+++ b/molecule/default/converge.yml
@@ -18,7 +18,6 @@
     - role: keycloak_quarkus
     - role: keycloak_realm
       keycloak_url: "{{ keycloak_quarkus_hostname }}"
-      keycloak_context: ''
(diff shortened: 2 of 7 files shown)


hcherukuri:
LGTM

PR #326, opened 2026-04-23
Add/example playbooks client scope auth flow

paulomenon:
The collection was missing modules for managing client scopes and authentication flows, forcing users to write raw uri calls against the Keycloak Admin REST API. This PR adds two new modules — keycloak_client_scope (manage client scopes with protocol mappers) and keycloak_authentication_flow (manage authentication flows with execution steps) — both built on top of the existing KeycloakAPI helper methods that were already in module_utils but never exposed as modules. Includes three example playbooks and documentation updates to the main README and the keycloak_realm role README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,15 @@ A requirement file is provided to install:
 
 <!--end roles_paths -->
 
+### Included modules
+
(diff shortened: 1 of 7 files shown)


hcherukuri:
LGTM

PR #328, opened 2026-04-24
Fix sanity and molecule tests

hcherukuri:


hcherukuri:
https://github.com/hcherukuri/keycloak/actions/runs/24893820837

Issue #346, opened 2026-06-04
Naming consistency for clientscope/client scope


felix-grzelka:
##### SUMMARY
There are different modules related to clientscopes.
The module `keycloak_clientscope` from community.general seems to have been renamed to `keycloak_client_scope`.
While I don't mind renaming that module, it should be consistent with other modules, such as `keycloak_clientscope_rolemappings`, `keycloak_clientscope_type`, etc.

It seems that in the doc strings clientsopes is sometimes written as "client scopes".

I think it is important to make this consistent, because clientscopes must not be confused with the scopes of a client (usually called roles, i think) and the roles of a clientscope (which are also sometimes called scope).

I found this very confusing while working on the `keycloak_clientscope_rolemappings` module.

Maybe it would be also helpful to add some general documentation to explain the terms and agree on a convention throughout the modules.

PR #348, opened 2026-06-04
Remove normalize_keycloak_url

### LINKED ISSUE #347: `normalize_keycloak_url` breaks URLs
##### SUMMARY
`normalize_keycloak_url` breaks URLs.
The `normalize_keycloak_url` method, which is applied in the keycloak.py helper to the baseurl property, removes a trailing `/auth`, which breaks our setup.
This is a breaking change compared to the current state of `community.general`, where this function does not exist.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible [core 2.18.6]
```

##### COLLECTION VERSION

```
middleware_automation.keycloak           3.0.7  
```

##### STEPS TO REPRODUCE
Configure Keycloak to listen on path `/auth`, then run any module with the correct url.

##### EXPECTED RESULTS
`normalize_keycloak_url` is not called and the URL is not changed.
Called Endpoint should be:
`svc.cluster.local/auth/realms/master/protocol/openid-connect/token`, when passing `auth_keycloak_url: svc.cluster.local/auth`

##### ACTUAL RESULTS
But instead the following Endpoint is called, resulting in a 404:
`svc.cluster.local/realms/master/protocol/openid-connect/token`

felix-grzelka:
This fixes #347

### DIFF
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -154,18 +154,6 @@ URL_AUTHZ_CUSTOM_POLICY = "{url}/admin/realms/{realm}/clients/{client_id}/authz/
 URL_AUTHZ_CUSTOM_POLICIES = "{url}/admin/realms/{realm}/clients/{client_id}/authz/resource-server/policy"
 
 
-def normalize_keycloak_url(url: str) -> str:
-    """Normalize Keycloak base URL for Admin REST API access.
-
-    Keycloak 17+ (Quarkus) exposes the API at the server root without an /auth prefix.
-    WildFly-based Keycloak used /auth as the context path. Trailing slashes are removed.
-    """
-    url = url.rstrip("/")
-    if url.endswith("/auth"):
-        return url[:-5]
-    return url
-
-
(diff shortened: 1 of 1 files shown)


plugins/module_utils/identity/keycloak/keycloak.py
hcherukuri: Need to change here too ` base_url = module_params["auth_keycloak_url"]`

hcherukuri (commented):
Can you make these changes, I think this should fix the issue

hcherukuri:
Thanks for the PR @felix-grzelka.

hcherukuri (approved):
LGTM

felix-grzelka:
Just wondering: @hcherukuri when will Version 3.0.8 be released, with this fix?

hcherukuri:
@felix-grzelka Lets resolve this and then release the collection https://github.com/ansible-middleware/keycloak/issues/346

hcherukuri:
Feel free to add something here https://github.com/ansible-middleware/keycloak/blob/main/README.md#configure-with-modules

PR #349, opened 2026-06-08
Consistent naming of client scope

### LINKED ISSUE #346: Naming consistency for clientscope/client scope

felix-grzelka:
This attempts to fix: #346 
All identifiers in the code (and file names) with the substring `clientscope` have been renamed to `client_scope` (except some ansible identifiers, which have been left as `client-scope`).
All occurrences of `client_scope`, `clientscope`, and `client-scope` in (doc) strings have been replaced with `client scope`.

Since I changed some parameter names, this will be a breaking change.


felix-grzelka:
Found something else that might be fixed together with his.
The function `delete_client_role_scope_from_client` for example uses a parameter called `client_scope_id` but its not a client scope but the id of a client which owns a role (or scope).
I think this is very confusing and would suggest to rename this and similar occurrences to something like `role_owner_client_id` and the other parameter to `target_client_id`.
Feel free to suggest an alternative name.

felix-grzelka:
Looks like this is becoming quite a big change (at least with lines of code changed).
I'm willing to work further on this and make it good.
But in the meantime, would it be possible to have a patch version release with the other PR, that's already merged into main?
I have downstream PRs waiting for that to be released and this one here might take some time to get it right.

hcherukuri:
@felix-grzelka I have released v3.0.8 which should contain previously merged PR's. Have a nice day :)

plugins/modules/keycloak_client.py
```diff
@@ -1346,7 +1346,7 @@ def main():
 
     # See if it already exists in Keycloak
     if cid is None:
-        before_client = kc.get_client_by_clientid(module.params.get("client_id"), realm=realm)
+        before_client = kc.get_client_by_client_idd(module.params.get("client_id"), realm=realm)
```
hcherukuri: There is a typo here before_client = kc.get_client_by_client_id(module.params.get("client_id"), realm=realm)

hcherukuri: This might be casuing issues with CI

felix-grzelka: Thanks for the hint.
and thanks for the release

hcherukuri (approved):
LGTM

PR #358, opened 2026-06-24
Removing from __future__ import annotations broke runtime type hints using Sequence and union syntax, causing sanity import failures and molecule module errors

hcherukuri:


RanabirChakraborty (approved):
LGTM

PR #363, opened 2026-07-01
fix(keycloak_quarkus): avoid leaking maven provider password in logs

### LINKED ISSUE #362: keycloak_quarkus: maven provider password leaked in cleartext (missing no_log on 3 provider tasks)
`keycloak_quarkus` logs the **full provider dict — including `maven.password` — in cleartext** for provider entries that use maven authentication. Three provider-looping tasks are missing `no_log`, so a private-registry credential (e.g. a GitHub Packages PAT) ends up in normal playbook stdout/CI logs.

hwo-wd:
## What

Adds the conditional `no_log` (already used by the maven download/copy tasks) to the three provider-looping tasks that currently print the full `item` — including `maven.password` — in cleartext:

- `roles/keycloak_quarkus/tasks/prereqs.yml` → `Validate providers`
- `roles/keycloak_quarkus/tasks/install.yml` → `Download custom providers via http`
- `roles/keycloak_quarkus/tasks/install.yml` → `Copy local providers`

Because Ansible echoes the loop `item` even for skipped iterations, these tasks leak a private-registry credential (e.g. a GitHub Packages PAT) at default verbosity, on every run.

## Fix

```yaml
no_log: "{{ item.maven.password is defined and item.maven.password | length > 0 | default(false) }}"
```

Only entries carrying a secret (maven `password`) are censored; `url`/`local_path` providers stay fully visible for debugging. This is the exact idiom the `Download custom providers to localhost using maven` and `Copy maven providers` tasks already use.

Closes #362

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/roles/keycloak_quarkus/tasks/install.yml
+++ b/roles/keycloak_quarkus/tasks/install.yml
@@ -246,6 +246,7 @@
   become: "{{ keycloak_quarkus_install_require_privilege_escalation | default(true) }}"
   loop: "{{ keycloak_quarkus_providers }}"
   when: item.url is defined and item.url | length > 0
+  no_log: "{{ item.maven.password is defined and item.maven.password | length > 0 | default(false) }}"
   notify: "{{ ['invalidate keycloak theme cache', 'rebuild keycloak config', 'restart keycloak'] if not item.restart is defined or item.restart else [] }}"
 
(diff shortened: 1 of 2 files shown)


RanabirChakraborty (approved):
LGTM

PR #364, opened 2026-07-03
Fix health check URL to include HTTP relative path

### LINKED ISSUE #299: HTTP relative path not reflected in the health_url
I am using proxy to forward keycloak traffic on /auth path to the keycloak backend , 
i have set  keycloak_quarkus_http_relative_path , however the  Wait until keycloak service becomes active check fails.

i see that health_url does not account for this,
  health_url: "{{ keycloak_quarkus_health_check_url | default(keycloak_quarkus_hostname ~ '/' ~ (keycloak_quarkus_health_check_url_path | default('realms/master/.well-known/openid-configuration'))) }}"

 can this be added dynamically instead ?

i noticed that this variable exists "keycloak_quarkus_health_check_url_path", but should it be used if i am defining keycloak_quarkus_http_relative_path  already ?

cihlamar:
- Include keycloak_quarkus_http_relative_path in health_url construction
- Use rstrip('/') to handle both default '/' and custom paths like '/auth'
- Add keycloak_quarkus_health_check_validate_certs option for self-signed certs

Fixes #299

### DIFF
--- a/roles/keycloak_quarkus/defaults/main.yml
+++ b/roles/keycloak_quarkus/defaults/main.yml
@@ -186,6 +186,9 @@ keycloak_quarkus_restart_strategy: restart/serial.yml
 keycloak_quarkus_restart_health_check: true
 keycloak_quarkus_restart_health_check_delay: 10
 keycloak_quarkus_restart_health_check_retries: 25
+keycloak_quarkus_validate_certs: true
+keycloak_quarkus_health_check_url_path: "realms/{{ keycloak_quarkus_master_realm }}/.well-known/openid-configuration"
+keycloak_quarkus_health_check_url: "{{ keycloak_quarkus_hostname ~ keycloak_quarkus_http_relative_path.rstrip('/') ~ '/' ~ keycloak_quarkus_health_check_url_path }}"
 keycloak_quarkus_restart_pause: 15
 
 keycloak_quarkus_force_install: false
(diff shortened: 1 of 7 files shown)


cihlamar:
Hi @RanabirChakraborty , please review.

roles/keycloak_quarkus/defaults/main.yml
```diff
@@ -188,4 +188,9 @@ keycloak_quarkus_restart_health_check_delay: 10
 keycloak_quarkus_restart_health_check_retries: 25
 keycloak_quarkus_restart_pause: 15
 
+# Health check configuration
+keycloak_quarkus_health_check_url_path: realms/master/.well-known/openid-configuration
```
RanabirChakraborty: `keycloak_quarkus_health_check_url_path` embeds master, which is a common hardening target to disable. If the master realm is disabled, health checks always 404 even when Keycloak is healthy. At minimum, the realm name should be a separate variable.

RanabirChakraborty (changes requested):
All looks good, just a small change needed.

roles/keycloak_quarkus/defaults/main.yml
cihlamar: done.

RanabirChakraborty (approved):
LGTM

PR #366, opened 2026-07-20
AMW-592 AMW Github Actions are failing because of latest molecule

RanabirChakraborty:
Issue: https://redhat.atlassian.net/browse/AMW-592

### DIFF
--- a/molecule/debian/molecule.yml
+++ b/molecule/debian/molecule.yml
@@ -6,7 +6,7 @@ platforms:
     image: ghcr.io/hspaans/molecule-containers:debian-13
     pre_build_image: true
     privileged: true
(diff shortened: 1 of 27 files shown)


hcherukuri (commented):
Is there a reason to add validate_certs: false in molecule tests?

RanabirChakraborty:
The ansible.builtin.uri module and open_url() now route even plain http:// connections through Python 3.12's SSL initialisation code path before the actual connection is made. This causes `[ASN1: NOT_ENOUGH_DATA] not enough data (_ssl.c:4192) `

hcherukuri (approved):

hcherukuri (approved):

Issue #367, opened 2026-07-23
community.general.keycloak_realm_users_info is missing in this collection


felixfontein:
##### SUMMARY
It was added in https://github.com/ansible-collections/community.general/pull/12105 and likely got forgotten during the migration.

I am in the process of deprecating all c.g keycloak modules and redirecting users here, and I noticed this collection doesn't have that module yet :)

##### ISSUE TYPE
 - Feature Idea

hcherukuri:
This is now merged, closing the issue.

felixfontein:
@hcherukuri do you already have plans when this is going to be released?

PR #372, opened 2026-07-24
Add community.general.keycloak_realm_users_info into keycloak collection

hcherukuri:
https://github.com/ansible-middleware/keycloak/issues/367

### DIFF
--- a/plugins/modules/keycloak_realm_users_info.py
+++ b/plugins/modules/keycloak_realm_users_info.py
@@ -22,10 +22,6 @@ description:
   - Authentication can be performed either with username/password or with a token.
   - The names of module options are snake_case versions of the camelCase ones found in the Keycloak API and its documentation
     at U(https://www.keycloak.org/docs-api/latest/rest-api/index.html).
-attributes:
-  action_group:
-    # Originally added in community.general 10.2.0
-    version_added: "3.1.0"
 
 options:
   realm:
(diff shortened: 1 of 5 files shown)


plugins/modules/keycloak_realm_users_info.py
```diff
@@ -0,0 +1,127 @@
...
+description:
+  - This module retrieves all users from a specified Keycloak realm using the Keycloak REST API.
+  - Access to the REST API is performed via OpenID Connect. The user and client used must have the necessary permissions.
+  - Authentication can be performed either with username/password or with a token.
+  - The names of module options are snake_case versions of the camelCase ones found in the Keycloak API and its documentation
+    at U(https://www.keycloak.org/docs-api/latest/rest-api/index.html).
+attributes:
+  action_group:
+    # Originally added in community.general 10.2.0
+    version_added: "3.1.0"
```
felixfontein: I'd remove this; since the module is added in 3.1.0, it's also clear that it wasn't added to the action group before that. This `version_added` was only there in community.general because we completely missed it during the review (it should never have had a version before the module's `version_added` anyway)
```suggestion
```

felixfontein (commented):
Thanks for working on this!

hcherukuri:
> Thanks for working on this!

Thanks, updated!

RanabirChakraborty (approved):
LGTM

PR #377, opened 2026-08-03
Add keycloak_organization and keycloak_organization_info modules

snecklifter:
## Summary

- Adds `keycloak_organization` module for CRUD management of Keycloak Organizations (Keycloak 26+), with full `check_mode`, `diff_mode`, and idempotency support
- Adds `keycloak_organization_info` module for listing/searching organizations by name with exact or partial matching
- Adds 6 API methods to the `KeycloakAPI` class and `URL_ORGANIZATIONS`/`URL_ORGANIZATION` endpoint constants
- Adds comprehensive Molecule integration tests covering create, idempotency, update, info queries, attributes, and delete flows

## Motivation

Keycloak 26 introduced Organizations as a stable feature for B2B multi-tenancy within a single realm. Currently there is no Ansible module support for organizations, requiring users to make direct REST API calls via `ansible.builtin.uri`. This PR adds declarative module support so organizations can be managed like other Keycloak resources (groups, clients, realms, etc.).

## Module parameters — `keycloak_organization`

| Parameter | Type | Description |
|-----------|------|-------------|
| `state` | str (present/absent) | Default: present |
| `realm` | str | Default: master |
| `id` | str | UUID for direct lookup |
| `name` | str | Required for create, used as lookup key |
| `alias` | str | URL-friendly, immutable after creation |
| `enabled` | bool | Disabled orgs block member auth |
| `description` | str | Free-text description |
| `domains` | list of dict(name, verified) | Associated internet domains |
| `attributes` | dict | Custom key/value metadata |

## Module parameters — `keycloak_organization_info`

| Parameter | Type | Description |
|-----------|------|-------------|
| `realm` | str | Required |
| `name` | str | Optional search filter |
| `exact` | bool | Exact vs partial match (default: true) |

## Test plan

- [x] Python compilation check passes for all modified files
- [x] `ansible-doc` parses both new modules correctly
- [x] `yamllint` passes with no errors
- [x] `ansible-lint` passes at production profile (only pre-existing `name[casing]` warnings)
- [ ] `molecule test -s keycloak_modules` — integration tests exercise:
  - Enable organizations on ephemeral realm
  - Create organization with name, description, enabled, domains
  - Idempotency check on create
  - Update organization description
  - Idempotency check on update
  - Query by name via info module
  - List all organizations via info module
  - Create organization with custom attributes
  - Delete organizations
  - Idempotency check on delete

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -155,6 +155,9 @@ URL_AUTHZ_RESOURCES = "{url}/admin/realms/{realm}/clients/{client_id}/authz/reso
 URL_AUTHZ_CUSTOM_POLICY = "{url}/admin/realms/{realm}/clients/{client_id}/authz/resource-server/policy/{policy_type}"
 URL_AUTHZ_CUSTOM_POLICIES = "{url}/admin/realms/{realm}/clients/{client_id}/authz/resource-server/policy"
 
+URL_ORGANIZATIONS = "{url}/admin/realms/{realm}/organizations"
+URL_ORGANIZATION = "{url}/admin/realms/{realm}/organizations/{id}"
+
 
 def keycloak_argument_spec() -> dict[str, t.Any]:
     """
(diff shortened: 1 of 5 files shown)


plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -3508,6 +3511,93 @@ def delete_client_role_scope_from_realm(self, roles, clientid, realm: str = "mas
 
         return self.get_client_role_scope_from_realm(clientid, realm)
 
+    def get_organizations(self, filter=None, realm: str = "master"):
```
RanabirChakraborty: filter is a Python built-in. Using it as a parameter name shadows it within the method. Rename to search_filter or query.

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -3508,6 +3511,93 @@ def delete_client_role_scope_from_realm(self, roles, clientid, realm: str = "mas
...
+            return self._request_and_deserialize(org_url, method="GET")
+        except HTTPError as e:
+            if e.code == HTTPStatus.NOT_FOUND:
+                return None
+            else:
+                self.fail_request(e, msg=f"Could not fetch organization {org_id} in realm {realm}: {e}")
+        except Exception as e:
+            self.module.fail_json(msg=f"Could not fetch organization {org_id} in realm {realm}: {e}")
+
+    def get_organization_by_name(self, name, realm: str = "master"):
```
RanabirChakraborty, line 3549: This method calls quote(name, safe='') but the diff doesn't show from urllib.parse import quote being added to this file. keycloak_organization_info.py imports it separately. If it's not already present in keycloak.py, this will raise a NameError at runtime. Please verify.

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -3508,6 +3511,93 @@ def delete_client_role_scope_from_realm(self, roles, clientid, realm: str = "mas
...
+        """
+        org_url = URL_ORGANIZATION.format(url=self.baseurl, realm=realm, id=org_id)
+        try:
+            return self._request_and_deserialize(org_url, method="GET")
+        except HTTPError as e:
+            if e.code == HTTPStatus.NOT_FOUND:
+                return None
+            else:
+                self.fail_request(e, msg=f"Could not fetch organization {org_id} in realm {realm}: {e}")
+        except Exception as e:
```
RanabirChakraborty, line 3546: why not use self.fail_request for consistency?

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -3508,6 +3511,93 @@ def delete_client_role_scope_from_realm(self, roles, clientid, realm: str = "mas
...
+        """Fetch an organization by name using search API with exact matching.
+        Returns None if not found.
+        :param name: Name of the organization to fetch.
+        :param realm: Realm in which the organization resides; default 'master'.
+        """
+        search_url = URL_ORGANIZATIONS.format(url=self.baseurl, realm=realm)
+        search_url = f"{search_url}?search={quote(name, safe='')}&exact=true"
+        try:
+            orgs = self._request_and_deserialize(search_url, method="GET")
+            for org in orgs:
```
RanabirChakraborty: The URL already sends exact=true to the server, so the API guarantees the result matches the name exactly. The loop seems to be unnecessary.

plugins/modules/keycloak_organization.py
```diff
@@ -0,0 +1,411 @@
+
+# Copyright (c) 2025, Chris Brown
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import (absolute_import, division, print_function)
```
RanabirChakraborty, line 6: I think this part of code is not pyhton 3.12+ compatible. please recheck.

plugins/modules/keycloak_organization.py
```diff
@@ -0,0 +1,411 @@
...
+
+    if before_org is None:
+        before_org = {}
+
+    normalise_org(before_org)
+
+    changeset = {}
+    for param in org_params:
+        new_param_value = module.params.get(param)
+        if param == "alias" and before_org:
```
RanabirChakraborty, line 329: need to add some warning, if it fails.

RanabirChakraborty (changes requested):

RanabirChakraborty:
@snecklifter Also the CI is failing, please verify what's causing the failure.

plugins/module_utils/identity/keycloak/keycloak.py
snecklifter: Good catch — renamed to `search_filter` to avoid shadowing the built-in.

plugins/module_utils/identity/keycloak/keycloak.py
snecklifter: `quote` is already imported at line 15 of `keycloak.py`: `from urllib.parse import quote, urlencode`. Verified it works at runtime.

plugins/module_utils/identity/keycloak/keycloak.py
snecklifter: Fixed — now using `self.fail_request` consistently in both `get_organization_by_id` and `get_organization_by_name`.

plugins/module_utils/identity/keycloak/keycloak.py
snecklifter: Agreed — simplified to `return orgs[0] if orgs else None` since `exact=true` guarantees the match.

plugins/modules/keycloak_organization.py
snecklifter: This is the same `from __future__ import` pattern used by every other module in the collection (e.g. `keycloak_group.py`, `keycloak_client.py`, `keycloak.py` itself). It's compatible with Python 3.12+.

plugins/modules/keycloak_organization.py
snecklifter: Added — the module now emits a warning via `module.warn()` when the user provides an alias that differs from the existing one, explaining that it is immutable and will be ignored for the update.

snecklifter:
All review feedback has been addressed in the latest push — renamed `filter` to `search_filter`, switched to `self.fail_request` consistently, simplified the name lookup, and added a `module.warn()` for immutable alias changes. Ready for re-review when you get a chance. Thanks @RanabirChakraborty!

RanabirChakraborty (approved):
LGTM

hcherukuri (approved):

Issue #382, opened 2026-08-13
keycloak_authentication_v2: Validation fails for FormAction providers


NilsBergmann:
##### SUMMARY

I'm trying to migrate the configuration of my flows to keycloak_authentication_v2. I'm running into an issue when configuring my registration flows: The module gets the list of supported authentication providers from Keycloak's API and validates the providers used, but it doesn't include FormAction providers (like the [default registration steps](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/authentication/forms/RegistrationPassword.java)), so it fails when attempting to use them. 

As it is I am unable to use the module for managing my registration flows and I'm stuck on v1.

Proposed solution: Also query "/admin/realms/{realm}/authentication/form-action-providers" for valid FormActions and support them as well.

I'll try and see if I can create a PR for this :)

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible [core 2.16.3]
  config file = None
  configured module search path = ['/home/nilsbergmann/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /home/nilsbergmann/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.12.3 (main, Jun 19 2026, 12:46:00) [GCC 13.3.0] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```
ansible-galaxy collection list

# /home/nilsbergmann/.ansible/collections/ansible_collections
Collection                               Version
---------------------------------------- -------
ansible.posix                            2.1.0  
community.docker                         5.0.5  
community.general                        11.4.9 
fedora.linux_system_roles                2.4.0  
middleware_automation.common             1.2.6  
middleware_automation.keycloak           3.0.10 

# /usr/lib/python3/dist-packages/ansible_collections
Collection                               Version
---------------------------------------- -------
amazon.aws                               7.2.0  
ansible.netcommon                        5.3.0  
ansible.posix                            1.5.4  
ansible.utils                            2.12.0 
ansible.windows                          2.2.0  
arista.eos                               6.2.2  
awx.awx                                  23.6.0 
azure.azcollection                       1.19.0 
check_point.mgmt                         5.2.2  
chocolatey.chocolatey                    1.5.1  
cisco.aci                                2.8.0  
[...]

PR #374, opened 2026-07-30
369 add integration tests for keycloak_authentication_v2

thomasbargetzrise:
This PR adds integration tests for the `keycloak_authentication_v2` module.

The tests are based on those from the community.general collection: https://github.com/ansible-collections/community.general/tree/main/tests/integration/targets/keycloak_authentication_v2

This PR includes **only** the integration tests for `keycloak_authentication_v2`.

Partially fixes #369

PR #384, opened 2026-08-14
authentication_v2: Fix validation for FormActions

NilsBergmann:
See #382.

This PR adds support for managing form actions with authentication_v2, previously the validation only allowed authenticator providers, which causes issues when attempting to configure form action flows like the registration flow.

I adjusted the validation to also fetch the installed form actions and made it differentiate between missing provider IDs and steps being misplaced in an invalid parent (sub-)flow.

Because there is a open PR modifying and restructuring the tests https://github.com/ansible-middleware/keycloak/pull/374 this is a draft for now, I'll update it to match the new test structure once that PR is merged

### DIFF
--- a/plugins/modules/keycloak_authentication_v2.py
+++ b/plugins/modules/keycloak_authentication_v2.py
@@ -812,33 +812,50 @@ def create_authentication_execution_spec_options(depth: int) -> dict[str, t.Any]
     return options
 
 
-def validate_executions(kc: KeycloakAPI, realm: str, executions: dict) -> None:
-    valid_providers = kc.get_authenticator_providers(realm)
-    valid_provider_ids = {provider["id"] for provider in valid_providers}
+def validate_executions(kc: KeycloakAPI, realm: str, executions: dict, flow_type: str) -> None:
+    valid_provider_ids_by_flow_type = {
+        "basic-flow": {provider["id"] for provider in kc.get_authenticator_providers(realm)},
+        "form-flow": {provider["id"] for provider in kc.get_form_action_providers(realm)},
+    }
+    all_valid_provider_ids = set().union(*valid_provider_ids_by_flow_type.values())
 
-    invalid_provider_ids = validate_executions_rec(valid_provider_ids, executions)
-    if len(invalid_provider_ids) > 0:
-        invalid_provider_ids_str = ", ".join(f"'{item}'" for item in invalid_provider_ids)
-        raise ValueError(
-            f"The following execution providerIds are unknown and therefore invalid: {invalid_provider_ids_str}"
+    invalid = validate_executions_rec(valid_provider_ids_by_flow_type, executions, flow_type)
+    if len(invalid) > 0:
+        unknown = sorted({provider_id for provider_id, _ in invalid if provider_id not in all_valid_provider_ids})
+        wrong_type = sorted(
+            {(provider_id, enclosing_flow_type) for provider_id, enclosing_flow_type in invalid if provider_id in all_valid_provider_ids}
         )
 
+        messages = []
+        if unknown:
+            messages.append("unknown providerIds: " + ", ".join(f"'{item}'" for item in unknown))
+        if wrong_type:
+            messages.append(
+                "providerIds not valid for their flow type: "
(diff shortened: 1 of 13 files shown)


plugins/modules/keycloak_authentication_v2.py
```diff
@@ -812,33 +812,50 @@ def create_authentication_execution_spec_options(depth: int) -> dict[str, t.Any]
...
+def validate_executions_rec(valid_provider_ids_by_flow_type: dict, executions: dict, flow_type: str) -> list:
+    invalid = []
+    valid_provider_ids = valid_provider_ids_by_flow_type.get(flow_type)
     for execution in executions:
         provider_id = execution["providerId"]
         sub_flow = execution["subFlow"]
         if provider_id is not None:
-            if provider_id not in valid_provider_ids:
-                invalid_provider_ids.append(provider_id)
+            if valid_provider_ids is not None and provider_id not in valid_provider_ids:
```
thomasbargetzrise: Why is the check `valid_provider_ids is not None` required?

plugins/modules/keycloak_authentication_v2.py
thomasbargetzrise, line 521: Please add yourself as an author

plugins/modules/keycloak_authentication_v2.py
NilsBergmann: It might be None if a flow type we do not support yet is being used and `valid_provider_ids_by_flow_type.get(flow_type)` returns None, in that case attempting `provider_id not in  valid_provider_ids` would raise a TypeError and end the validation early. 

But explicitly checking for that case and returning a matching error message `"Unspported flow type '{}'"` would be better than this, I'll take a look into it

RanabirChakraborty:
@NilsBergmann What's the status on the PR?

NilsBergmann:
> @NilsBergmann What's the status on the PR?

I got swamped by my day to day workload and this got pushed back, sorry for that. I'll try to make time to complete this in the next days

open tasks are

- Improve handling of unknown flow types (see above discussion)
- Match test structure introduced in https://github.com/ansible-middleware/keycloak/pull/374

NilsBergmann:
I updated the validation logic to more explicitly handle the described cases and simplified it a bit.
I also added the handling of client-flow creation, which, just like form-flow, wasn't covered before.

I also adjusted the test cases to match the new structure and added a new test case for client-flows

RanabirChakraborty:
@NilsBergmann Can you please check why the specific molecule scenario is failing?
and also if there are similar commits then merge them into one.

NilsBergmann:
The issue seems to be the providerId of the form-flow, which defaults to 'registration-page-form' and is not set on our expected-object for the comparison. This causes the test to fail due to an alleged diff when rerunning the task for the idempotency check.

If I understand Keycloaks source code correctly: Unlike basic-flow and client-flow, form-flow has a SPI for the renderer. Keycloak ships with only 'registration-page-form', the id hardcoded in https://github.com/ansible-middleware/keycloak/blob/8d5526aeb6575fda19934ada739a9bb73dbcca3d/plugins/module_utils/identity/keycloak/keycloak.py#L2490

The Keycloak admin UI defaults to the same ID when creating a form-flow, but there is code for a selection dropdown that's hidden if only one exists https://github.com/keycloak/keycloak/blob/8fb918b250b3e52ea16e5e4adec08d38429a5b8a/js/apps/admin-ui/src/authentication/components/modals/AddSubFlowModal.tsx#L54.

A custom Keycloak deployment could theoretically register more than one form-flow-rendering provider through FormAuthenticatorFactory.

In my opinion there are two possible solutions here:
- Exclude providerId from the diff comparison, which I've done for now
- Expand the module to allow setting custom providerIds for flows, including validation through getting the available FormAuthenticatorFactorys and preventing setting a provider for non-form-flows.

Option 2 seems to be out of scope for this PR, what is your opinion on that?

I'll clean up the history once the tests pass :)

RanabirChakraborty:
@NilsBergmann testcases are passing fine now, please make it a single commit now.

RanabirChakraborty (approved):
LGTM

PR #392, opened 2026-08-21
Normalize hostname that includes http_relative_path suffix

### LINKED ISSUE #388: `http-relative_path` follow up: inconsistencies between `keycloak_quarkus_hostname` and keycloak health check
@cihlamar has just recently [committed](https://github.com/ansible-middleware/keycloak/commit/4f62617f2cf34fb16fbf6501da0d756d1e50184b) a fix for #299.

While #299 is fixed such that `http_relative_path` is included in health checks now, this fix breaks different things, especially for instances which were previously RH-SSO and have been migrated to RHBK since; due to this legacy, they usually include `/auth` such that JBoss could route properly; this artifact stuck with RHBK instances since it's part of every external federation etc., so not an easy task to get rid of.

KC introduced hostname v2 feature a while back and since quite a while I've configured my playbook as follows:

```
rhbk_hostname: "https://sso-test.example.local/auth"
rhbk_http_relative_path: "/auth" # must start with `/`
```

this would result in `conf/keycloak.conf` being populated as a (working!) configuration:

```text
# HTTP
...
http-relative-path=/auth
...
# Client URL configuration
hostname=https://sso-test.example.local/auth
...
```

and the health check would be run against `https://sso-test.example.local/auth/realms/master/.well-known/openid-configuration`.

Now with the fix provided by @cihlamar, things are templated differently:

```diff
- rhbk_hostname: "https://sso-test.example.local/auth"
+ rhbk_hostname: "https://sso-test.example.local"
rhbk_http_relative_path: "/auth" # must start with `/`
```

```diff
# HTTP
...
http-relative-path=/auth
...
# Client URL configuration
- hostname=https://sso-test.example.local/auth
+ hostname=https://sso-test.example.local
...
```

This has the effect that:
- hostname is wrong: it's not the fully qualified domain expected by hostname v2; the UI will thus no longer work as expected
- health check passes just fine now since this maps to the correct url
- suffixing `/auth` to hostname leads to:
  - a working ui
  - failing health checks due to dupicate `/auth/auth/...` URI

I'm writing this to kick of a discussion on how to best proceed here: I'm arguing in favor of keeping `hostname` pure w/o a lot of extra templating (Edit: i.e., not fiddling with `http_relative_path` in `keycloak_quarkus_hostname` whatsoever) since it easily translates to raw keycloak configuration, but I'd be open for discussions about adding `http_relative_path` to the `keycloak.conf` template.
Any thoughts, @RanabirChakraborty, @cihlamar?

cihlamar:
Strip duplicate context path from hostname when it already contains the http_relative_path value (e.g. /auth). Prevents double path in both keycloak.conf and health check URL for RH-SSO to RHBK migrations.

Fixes #388

[AMW-612](https://redhat.atlassian.net/browse/AMW-612)


hwo-wd:
Thanks for the pull request. This does not fix #388  though, at least not without a deprecation warning: as stated in 388, if I specify `keycloak_quarkus_hostname` w/o the `/auth` suffix, the UI does no longer work.
But v3.0.11 is breaking these anyway, but I guess upstream goes through separate testing and this thus won't suffer from it...

As a reviewer, I'd also be missing tests.

Also, I'm sad that no discussion happened just yet ;-)

> I'm arguing in favor of keeping hostname pure w/o a lot of extra templating (Edit: i.e., not fiddling with `http_relative_path` in `keycloak_quarkus_hostname` whatsoever) since it easily translates to raw keycloak configuration, but I'd be open for discussions about adding http_relative_path to the keycloak.conf template.

cihlamar:
> Thanks for the pull request. This does not fix #388 though, at least not without a deprecation warning: as stated in 388, if I specify `keycloak_quarkus_hostname` w/o the `/auth` suffix, the UI does no longer work. But v3.0.11 is breaking these anyway, but I guess upstream goes through separate testing and this thus won't suffer from it...
> 
> As a reviewer, I'd also be missing tests.
> 
> Also, I'm sad that no discussion happened just yet ;-)
> 
> > I'm arguing in favor of keeping hostname pure w/o a lot of extra templating (Edit: i.e., not fiddling with `http_relative_path` in `keycloak_quarkus_hostname` whatsoever) since it easily translates to raw keycloak configuration, but I'd be open for discussions about adding http_relative_path to the keycloak.conf template.

Hi @hwo-wd, ok, reworked, now the hostname is untouched, and just relative path is removed from the hostname when composing the final URL. Perhaps this would work better this time. Test is included.

RanabirChakraborty:
@hwo-wd are you okay with the fix now?

hwo-wd (approved):
lgtm, thx

RanabirChakraborty (approved):

RanabirChakraborty:
Thanks @hwo-wd @cihlamar

PR #408, opened 2026-09-06
Add use_netrc option to prevent .netrc overriding bearer token

cihlamar:
When ~/.netrc contains an entry matching the Keycloak host, open_url() silently replaces the OIDC Bearer header with Basic auth, causing 401 errors. Add use_netrc option (default false) to all keycloak modules and pass it through to both open_url() call sites.

Includes molecule test scenario (netrc) that verifies both behaviors:
- default use_netrc=false ignores .netrc (fix works)
- use_netrc=true picks up .netrc and gets 401 (proves the bug)

[AMW-625](https://redhat.atlassian.net/browse/AMW-625)


RanabirChakraborty (approved):
LGTM

PR #419, opened 2026-09-14
Add support for the `description` attribute of keycloak groups

nwintering:
Previously it was not possible to set the description of groups in Keycloak via the `keycloak_group`-ansible-role.
Since this attribute is readily available in the Keycloak API, adding support is only a matter of adding a module parameter and corresponding documentation.


nwintering:
I am not sure if there is anything I can do about the failed checks. Looks to my like there is some issue with CI that causes it to time out?

hcherukuri (approved):
LGTM

Issue #423, opened 2026-09-17
keycloak_quarkus_health_check_url default crashes with AttributeError when keycloak_quarkus_hostname is unset


idNoRD:
##### SUMMARY

keycloak_quarkus_health_check_url's default calls .rstrip('/') directly on keycloak_quarkus_hostname, which defaults to null. Crashes with AttributeError when hostname is left unset (a documented, valid config when using dynamic hostname resolution). Regression from commits 4f62617f2c and b48b0b7334; pre-regression default was simply "".

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION
```
ansible --version
ansible [core 2.21.4]
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/ansible/lib64/python3.12/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = ./ansible
  python version = 3.12.13 (main, Jul 15 2026, 00:00:00) [GCC 11.5.0 20240719 (Red Hat 11.5.0-5)] (/opt/ansible/bin/python3.12)
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)
```

##### COLLECTION VERSION

```
middleware_automation.common             1.2.6  
middleware_automation.keycloak           3.1.0 
```

##### STEPS TO REPRODUCE
```yaml
---
- hosts: localhost
  become: true
  vars:
    keycloak_quarkus_bootstrap_admin_user: admin
    keycloak_quarkus_bootstrap_admin_password: changeme
    keycloak_quarkus_hostname_strict: false
    keycloak_quarkus_health_check_url: "http://localhost:9080/auth/realms/master/.well-known/openid-configuration"
    # keycloak_quarkus_hostname intentionally unset
  tasks:
    - include_role:
        name: middleware_automation.keycloak.keycloak_quarkus
```

##### EXPECTED RESULTS
Role runs normally. keycloak_quarkus_hostname being unset shouldn't matter — keycloak_quarkus_health_check_url is already supplied explicitly.

##### ACTUAL RESULTS
```
[ERROR]: Task failed: Finalization of task args for 'ansible.builtin.validate_argument_spec' failed:
Error while resolving value for 'argument_spec': object of type 'NoneType' has no attribute 'rstrip'

[...]

PR #424, opened 2026-09-17
Fix keycloak_quarkus_health_check_url crash when hostname is unset

cihlamar:
Default keycloak_quarkus_hostname to empty string instead of null to prevent AttributeError when .rstrip() is called on NoneType in the health_check_url default expression.

Jira: [AMW-634](https://redhat.atlassian.net/browse/AMW-634)
GH Issue: #423


hcherukuri (approved):
LGTM

